PR #2, opened 2024-11-05
Add runtime config and production docker file (GSI-1109)

Cito:
- add support for runtime configuration via YAML files and ENV vars
- the config settings are injected directly into `index.html` to avoid an additional request
- add `run.js` script for running in development and production mode
- run on port 8080 in both cases
- add production docker file using the static web server package
- removes polyfills for Zone.js that were forgotten in #1


.devcontainer/data-portal.yaml
```diff
@@ -0,0 +1,7 @@
+# development configuration
+
+host: 127.0.0.1
+port: 8080
+
+mass_url: /api/mass
+metldata_url: /api/metldata
```
ac-jorellanaf, line 7: We need to set up editoconfig for these small issues of missing end newlines

Cito: We will add pre-commit hooks for checking this and many other things.

Cito: The new-lines at the end of files should be taken care of by prettier. We need to add the prettier plugin, a suitable config and a commit hook for prettier. This is part of GSI-1104.

ac-jorellanaf: Well, then our two options for this issue are disabled then 😄

ac-jorellanaf (approved):
All works well!

SilverLinings89 (approved):
LGTM

PR #4, opened 2024-11-06
Feature eslint

SilverLinings89:
Work on #GSI-1104

Contains a husky and eslint config that uses prettier recommended, angular eslint recommended, contains fixes for previous linter errors and some explanation in the readme.


SilverLinings89:
I need to add some more compodoc rules and a readme entry about compodoc.

package.json
```diff
@@ -6,10 +6,20 @@
...
+    "test": "ng test",
+    "lint": "ng lint",
+    "lf": "ng lint --fix",
+    "prepare": "husky",
+    "docs": "compodoc -p tsconfig.doc.json -s -r 8081",
+    "compodoc:build": "compodoc -p tsconfig.doc.json",
+    "compodoc:build-and-serve": "compodoc -p tsconfig.doc.json -s -r 8081",
+    "compodoc:serve": "compodoc -s"
   },
+  "type": "module",
```
Cito, line 18: As Jordy also noticed, this breaks the `run.js` script. It must be modified to use the `import` syntax, and needs some more changes because it uses `__dirname`.

Cito: According to the [Angular docs](https://angular.dev/tools/libraries/angular-package-format#esm-declaration), this should have been set already. Strange that `ng new` did not do this.

Cito (changes requested):
Overall, looks good. Made some suggestions.

package.json
SilverLinings89: ah sorry, I wasn't aware of the separate run.js file. Should have migrated that, too. Will do.
I think we should do our best to stay away from CommonJs. It's inferior in some technical aspects and codebases typically struggle if the standards are mixed. It was pretty annoying to get some of the eslint stuff working and it was basically only possible via esm.

Cito: I agree - ES modules are the future and the Angular docs also seem to recommend them.

README.md
```diff
@@ -57,3 +57,13 @@ Angular CLI does not come with an end-to-end testing framework by default. You c
...
+
+To ensure deterministic behavior, the pre-commit hook *does not* attempt to fix linter errors. Most of the time, you will be fine by simply running `ng lint --fix` which attempts to automatically fix most of the issues. If we ran that in the hook, however, you would be committing different code than the one you checked. So if you cannot commit your code, run lint fix. If that doesn't resolve all the issues (which you can see by running ng lint) resolve those issues and try again.
+
+## Ease of use
+
+For comfort, we are adding these shorthands: `npm run lint`, `npm run lf`` (for `lint --fix`) and `npm run docs` (to build and serve the documentation). Apart from seeing the linter warnings when you (try to) commit or run the linter manually, your IDE should also show you these warnings in the code and fixing (the auto-fixable ones) should be offered in the context menu on hover or via Ctrl-..
```
Cito, line 69:
```suggestion
For comfort, we are adding these shorthands: `npm run lint`, `npm run lf` (for `lint --fix`) and `npm run docs` (to build and serve the documentation). Apart from seeing the linter warnings when you (try to) commit or run the linter manually, your IDE should also show you these warnings in the code and fixing (the auto-fixable ones) should be offered in the context menu on hover or via Ctrl-..
```

Sorry, there was one backtick too much.

Cito (approved):

Cito (approved):

PR #7, opened 2024-11-08
Enforce the architecture matrix (GSI-1105)

Cito:
This PR enforces the architecture matrix as outlined in [ADR018](https://github.com/ghga-de/adrs/blob/main/docs/adrs/adr018_frontend_architecture.md).

The subdirectories for the vertical slices and layers should be added when they are needed. I did not include empty directories to avoid cluttering the repo with README or .gitkeep files. The purpose of the individual subdirectories is already explained in the main README.


README.md
```diff
@@ -76,4 +76,31 @@ To ensure deterministic behavior, the pre-commit hook *does not* attempt to fix
...
+| features | features | features        | ... | features | features |
+| ui       | ui       | ui              | ... | ui       | ui       |
+| services | services | services        | ... | services | services |
+| models   | models   | models          | ... | models   | models   |
+| utils    | utils    | utils           | ... | utils    | utils    |
+
+To create a clean architecture, the following rules are checked when importing modules from the architecture matrix:
+
+- Modules within a vertical slice must only import modules from the same slice, as these slices represent bounded contexts.
+- An exception is that all vertical slices are allowed to use modules from the *shared* vertical slice.
```
SilverLinings89, line 102: One idea that is missing is to force stuff out of shared if possible. Currently, there is no statement that we won't just move everything into shared. Do we have a definition of when that's acceptable?

SilverLinings89 (approved):
Looks good, just the one suggestion

README.md
Cito: It's hard to create rules for this. It should be the task of the reviewers to check that features are put into the right vertical slice and that a new vertical slice is added when a new feature area (bounded context) is introduced.

PR #8, opened 2024-11-11
GitHub Action for the linter

SilverLinings89:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,23 @@
+name: Eslint
```
Cito, line 1:
```suggestion
name: ESLint
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        node-version: [22]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Install pnpm
```
Cito, line 13: I would have expected that you first need to install Node.js and then pnpm, but it seems to work that way, too. So it's probably fine.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,23 @@
+name: Eslint
+on:
+  push:
+
+jobs:
+  build:
+    runs-on: ubuntu-24.04
+    strategy:
```
Cito, line 8: Would be easier to just specify the node version directly in the setup-node action and remove the matrix strategy. We don't need to run this on different node versions.

Cito (approved):
Looks good. Made some minor suggestions.

You also need to recreate the lock file.

.github/workflows/lint.yml
SilverLinings89: I used the recommended code from the pnpm documentation for this setup. I think I would keep it like that for now and update it if we notice any problems if that's OK

Cito: It should no cause any problems, it's just an unnecessary complication. But I'm also fine with keeping it like that.

PR #9, opened 2024-11-11
Tailwind support

SilverLinings89:
This also contains a  prettier plugin for sorting the class names. So tailwind class names will always be sorted the same way. This can by done by automatically lint-fixing. Otherwise, it will show a warning when running the linter.

As a consequence, from now on, there should be as little actual css as possible because most stuff can be done with tailwind and this removes the need for stylesheets and css in general while also providing optimally implemented rules and tree shaking without code duplication.


tailwind.config.js
Cito, line 1: Don't we also need a postcss.config.js file?

Cito (approved):
Have tested running development and production builds, works as expected.

Nice how seamless this integration using PostCSS works.

See also my two comments below.

tailwind.config.js
SilverLinings89: I don't think so. We only use defaults from what I can tell (meaning tailwind postcss defaults)

Cito: I was wondering because it's mentioned here as step 2: https://tailwindcss.com/docs/installation/using-postcss

PR #10, opened 2024-11-11
Support two different modes of running the dev server

Cito:
This PR improves the `dev_launcher` script to allow switching between "localhost" (MSW) and "staging" (real backend) by just passing the "staging" option on the command line.


Cito:
> Some comments: cert.crl should probably be gitignored? We probably should add some documentation on setting up the certificate issuer to avoid browsers rejecting access to the page.

Where do you get a crl file? I will merge this already and then create another PR for better documentation and basic auth support.

PR #16, opened 2024-11-19
Move runtime config to separate file

Cito:
I had injected the runtime configuration into the index.html file before in order to avoid the overhead of fetching an additional file from the server. However, since the configuration changes between the different development modes, this would cause unnecessary changes of the index file which is part of the repository. Removing the file from the repository and copying it from a template instead would have been an option. However, the index file is part of the Angular boilerplate that is managed by ng, therefore I decided to better keep it in the repository. So I reverted back to a very simple and clean solution that loads the configuration from a separate file, even if this is not the most efficient.

This PR also adapts the production Dockerfile to this change and to the use of pnmp instead of npm.

### DIFF
--- a/.devcontainer/dev_install
+++ b/.devcontainer/dev_install
@@ -3,8 +3,8 @@
 
 cd /workspace
 
-corepack prepare pnpm@9.12.3 --activate
-echo "Updating npm..."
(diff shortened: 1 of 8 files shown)


SilverLinings89 (approved):
LGTM

PR #17, opened 2024-11-20
Add authentication (GSI-1113)

Cito:
This PR implements the complete 2FA authentication flow in a rudimentary way.

To be done in follow-up PRs (see also TODO markers in the code):
- Make this work with MSW when this has been integrated, Currently, it only works with the staging backend.
- Add proper styling to the UI.
- Add success and error messages where indicated (preferably as toast messages).
- Add "can activate" guards for verification of the proper session state.
- Add "can deactivate" guards for making sure that the user "stays in the flow".
- Check edge cases and multi-tab browsing.


src/app/auth/features/confirm-totp/confirm-totp.component.scss
```diff
@@ -0,0 +1 @@
+// TODO: remove if not needed
```
ac-jorellanaf, line 1: Just delete it. Can be remade later 👍

ac-jorellanaf: Same for the others

src/app/portal/features/show-session/show-session.component.html
ac-jorellanaf, line 1: Should we even keep this?

ac-jorellanaf (approved):
Works well.
Minor comments but approved. 👍

src/app/auth/features/confirm-totp/confirm-totp.component.scss
Cito: Removed now. Generally, we should rarely need individual css thanks to tailwind.

src/app/portal/features/show-session/show-session.component.html
Cito: Let's remove it at then end when the base app is finished. It can be handy when working on the remaining issues (see bullet list in the description).

PR #18, opened 2024-11-26
More layout and styling (GSI-1180)

ac-jorellanaf:
Added icons (Material and FA Brands)
Added some additional styling to header and footer links.
Footer links remain to be wrapped in a button of some sort, but angular material limits button sizes, so custom CSS code needs adding.


src/app/portal/features/site-header/site-header.component.html
```diff
@@ -1,9 +1,20 @@
...
   <mat-nav-list>
-    <a routerLink="/" class="mx-1">Home</a
-    ><a routerLink="/browse" class="mx-1">Browse</a></mat-nav-list
+    <a
+      [id]="route === '/' ? 'activePage' : ''"
+      mat-button
+      routerLink="/"
+      class="homeLink mx-1"
+      >Home</a
+    ><a
```
Cito, line 11: I wonder whether this strange formatting of prettier is a bug or a feature?

Cito (commented):
Header and Footer look good already.

Some colors are a bit off in comparision with the legacy app.

For instance the active header button is brownish instead of orange, and the footer icons should be more white.

The larger footer icons need some kind of hover effect since they are not Material buttons.

The social media buttons should be more grey (as in the legacy app). You could also add a hover effect (e.g. changing the color to the current one).

Cito:
Can you also add the GHGA logo and the curved contour of the footer back? I think that was nice.

src/app/portal/features/site-header/site-header.component.html
ac-jorellanaf: Might be feature, to show that there are more tags after the closing tag?

Cito (changes requested):
- the footer is now at the bottom, but the page height is a bit greater than the window page (a scrollbar appears) and the links at the very bottom are not visible, even though the page content is not tall
- also, since the footer has "position: absolute" it can overlap page content when it is tall
- the background color of the "home" button is still brownish rgb(136, 32, 0) instead of orange rgb(232, 70, 20)
- the logo in the footer is stretched or compressed when the window width changes
- the text and icon color in the footer is still greenish rgb(229, 231, 235) instead of white

ac-jorellanaf:
> * the footer is now at the bottom, but the page height is a bit greater than the window page (a scrollbar appears) and the links at the very bottom are not visible, even though the page content is not tall
> * also, since the footer has "position: absolute" it can overlap page content when it is tall
> * the background color of the "home" button is still brownish rgb(136, 32, 0) instead of orange rgb(232, 70, 20)
> * the logo in the footer is stretched or compressed when the window width changes
> * the text and icon color in the footer is still greenish rgb(229, 231, 235) instead of white

The issue with the colours are due to the Material palette generation. Unless we manually force the palette colours as we wish, then we use the material generated ones

Cito (changes requested):
Footer positioning looks good to me now.

Regarding the colors, I guess we will then need to use a custom palette that matches the corporate design colors instead of generating it. But the [generator](https://github.com/angular/components/blob/main/src/material/schematics/ng-generate/theme-color/README.md) also accept further options for secondary and tertiary colors. Maybe that would be sufficient?

ac-jorellanaf:
> Footer positioning looks good to me now.
> 
> Regarding the colors, I guess we will then need to use a custom palette that matches the corporate design colors instead of generating it. But the [generator](https://github.com/angular/components/blob/main/src/material/schematics/ng-generate/theme-color/README.md) also accept further options for secondary and tertiary colors. Maybe that would be sufficient?

I already set it up that way, generating with our actual brand colours. We could otherwise just overwrite the colour variables ourselves.

```
// This file was generated by running 'ng generate @angular/material:theme-color'.
// Proceed with caution if making changes to this file.

...

// Note: Color palettes are generated from primary: #00393f, secondary: #e84614, tertiary: #cfe7cd, neutral: #005a5a```

Cito (commented):
Suggestion: Since the color issues seem to be more complicated due to the ng theme generator not working as expected, let's ignore them in this PR and fix them in a separate one.

Cito (changes requested):
The following things still look different compared with the legacy app:

- the top/bottom logo should have a vertical/horizontal bar and "DATA PORTAL" text
- the bottom logo should be much smaller
- the other icons in the footer should be a little bit smaller
- when hovering over the links in the footer, the color should change

The colors can be fixed in a separate PR.

Cito (approved):

PR #21, opened 2024-12-04
Add support for MSW

Cito:
This PR adds support for MSW for local testing without backend.

- We are following the same principle as in the legacy app.
- Mock responses must be specified as static data - this avoids intricate re-implementation of the real backend that will deviate from reality anyway.
- The responses are auto-converted to a list of handlers as expected by MSW.
- I have added two demo responses and a demo usage in the home page component that should be removed later.
- Support for authentication (needs some extra logic) will be done in a follow-up PR.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -28,6 +28,11 @@
                 "glob": "**/*",
                 "input": "public"
               },
+              {
+                "glob": "mockServiceWorker.js",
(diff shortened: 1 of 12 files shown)


ac-jorellanaf (approved):
LGTM

PR #24, opened 2024-12-13
Pin devcontainer base and fix issue with Node 22.12

Cito:
- The base docker image for the devcontainer was not pinned to an exact version, which means developers could end up using node 22.9 or node 22.12 depending on the time they built the devcontainer image.
- node 22.12 introduced the experimental require(esm) feature which throws an error when the http proxy middleware loads the configuration from proxy.conf.js.
- This PR pins the devcontainer base image to version 1.1.8 which uses node 22.12.
- Therefore this PR also explicitly disables the require(esm) feature using the `NODE_OPTIONS` env var.
- This PR also includes minor updates to pnpm and Angular.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,16 +1,19 @@
-FROM mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm
+FROM mcr.microsoft.com/devcontainers/typescript-node:1.1.8-22-bookworm
 
 RUN apt update && apt install -y dnsutils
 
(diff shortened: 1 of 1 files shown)


SilverLinings89 (approved):
We will eventually have to figure out a way to get around this flag (maybe we can benefit from the new feature) but as long as we have no real need for this feature we can put that on hold. LGTM

Cito:
As far as I understand the problem is that the proxy middleware simply doesn't support it yet, but I assume it will do so in the future as people complain. We can then enable the feature again.

PR #25, opened 2024-12-16
Add license and copyright headers to all ts files and the eslint config.

SilverLinings89:


eslint.config.js
```diff
@@ -47,6 +49,13 @@ export default [
         'error',
         { type: 'element', prefix: 'app', style: 'kebab-case' },
       ],
+      'header/header': [
+        2,
+        'block',
+        ['*', ' * @license Apache-2.0', ' * @copyright The GHGA Authors', ' '],
```
Cito, line 55: Is there a reason why you put the license first? Usually the copyright comes first, then the license.

Cito: Also, shouldn't we use `@license SPDX-License-Identifier: Apache-2.0` for better machine readability?

src/app/app.routes.ts
```diff
@@ -1,18 +1,24 @@
...
+ */
+
 import { inject } from '@angular/core';
 import { Routes } from '@angular/router';
 import { AuthService } from '@app/auth/services/auth.service';
 
 export const routes: Routes = [
   {
     path: '',
+    // eslint-disable-next-line jsdoc/require-jsdoc
```
Cito, line 13: Not in the scope of this PR, but is it possible/reasonable to disable the rule for all fat arrow functions (or at least for those without code blocks)?

Cito (approved):
See two suggestions/questions below.

eslint.config.js
SilverLinings89: For licenses I think that's not recommended because @license already specifies that matter. If I add it, JSDoc throws this error: Invalid JSDoc @license: "SPDX-License-Identifier: Apache-2.0"; expected SPDX expression: https://spdx.org/licenses/
But I will swap the order

src/app/app.routes.ts
SilverLinings89: I have fixed this in the latests commit. It is defined in the ESLint Config (ArrowFunctionExpression: false - used to be true)

eslint.config.js
Cito: The SPDX tag should probably also not go into a block comment, but the line must start with `/*` or `//`

https://spdx.dev/learn/handling-license-info/#how

SilverLinings89:
I have turned the comments into line comments

eslint.config.js
SilverLinings89: https://jsdoc.app/tags-license hier verwenden sie genau das Beispiel auf der JSDoc Seite als Block Kommentar, daran hatte ich mich orientiert. Aber wir können das machen wie wir wollen denke ich, weil ich vermute, dass die ganzen Parser wahrscheinlich sowieso nach den einzelnen Optionen in der SPDX List suchen.

ac-jorellanaf (approved):

Cito (approved):

eslint.config.js
Cito: Just for the records: [JSDoc](https://jsdoc.app/tags-license) uses block comments, [SPDX](https://spdx.dev/learn/handling-license-info/#how) tags use line comments, and they use different tags (`@license` vs `SPDX-License-Identifier:`).

This makes it impossible to combine them in one line.

We could add both types of license tags in separate lines, but this would be ugly and redundant. So we decided to include the JSDoc style license tag only.

PR #27, opened 2024-12-17
Add a description to all TypeScript modules

Cito:


Cito (commented):
This PR allows and requires a single-line description in the comment header of all TypeScript modules.

SilverLinings89 (approved):
LGTM. As a remark: I had previously not included such a line because I thought the goal was to be able to "blindly scroll past" the initial comment in the document. If we want to include information that is relevant in these headers, this is a great solution.

PR #28, opened 2024-12-18
Support more modes for running the dev server (GSI-1233)

Cito:
This PR makes the `run.js` script a bit more flexible. Besides running the production server, it can now run the dev server in four different modes:
- with mock API and fake authentication
- with real backend, but fake authentication
- with real OIDC provider, but mock backend
- with real backend and OIDC provider

The dev server is activated by passing the `--dev` option. The `dev_launcher` script calls the `run.js` script with this option, while passing on any other options.

See also the update documentation in the README file.

### DIFF
--- a/.devcontainer/create_cert
+++ b/.devcontainer/create_cert
@@ -2,18 +2,24 @@
 
 # create self-signed certificate for testing with the browser
 
-BASE_URL=${BASE_URL:-https://data.staging.ghga.dev}
-HOST=${BASE_URL#https://}
 CERTFILE=cert.pem
 KEYFILE=key.pem
 
 cd /workspace/.devcontainer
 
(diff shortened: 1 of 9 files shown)


ac-jorellanaf (changes requested):
Everything but `--with-oidc` alone works. It tries to use the localhost instead of the correct staging host. Everything else works well

Cito:
> Everything but `--with-oidc` alone works. It tries to use the localhost instead of the correct staging host.

It works for me. The env variable `data_portal_base_url` should not be set or set to `https://data.staging.ghga.dev` when testing. When I browse to https://data.staging.ghga.dev, I can login via LS Login, but the summary data is mocked.

ac-jorellanaf:
It works with the `data_portal_base_url` variable set to staging, but does not work when just unset.

ac-jorellanaf (approved):

PR #29, opened 2024-12-20
Use Jest instead of Jasmine and Karma for unit testing (GSI-1115)

Cito:
This PR adds tooling for using Jest instead of Jasmine and Karma that is currently still used by Angular but will be phased out. The README has been updated with the corresponding documentation.

Note that we are running the tests serially by default, because running them in parallel takes twice as much time (at least on my machine). One possible reason could be Jest compiling the Angular framework multiple times in this case. Did not spend much time trying to fix this, because we know that the Angular team will provide an official unit testing solution probably based on Jest soon, with this problem then hopefully already solved.

The examples show the use of normal query selectors and selectors like `getByRole` using `@testing-libraries/angular`. Using such selectors abstracts from implementation details and simulates the user interaction much better.

While testing I noticed that in some cases `@testing-libraries/dom` failed when selecting roles or checking visibility, because there are tailwind classes that contain special characters like "base-1/4" and the library for some reason converted these to a selector without escaping the special characters. I could not reproduce it in a simpler setup to create a bug report, but I hope the problem rarely occurs or will be fixed in future versions. We can always use normal query selectors.

Since we are already using ESLint, we could also run the unit tests as part of this process using jest-runner-eslint, but it makes more sense to keep them separate since linting and testing has a different focus and we don't want the unit tests to run as  part of the pre-commit checks.


SilverLinings89:
On my machine, running the tests in parallel takes the same time as sequential execution - maybe 10 percent more. Just as a data-point. I also see the deprecation warning 10 times (once per suite) rather than only once, indicating that each thread performs the same task. I agree that likely this has to do with compilation work being repeated and it makes no sense to waste time on going from 5 seconds to 3 on this.

Cito:
Ok, looks like it is less bad with a more modern CPU with many cores. But I think using the serial mode as default is still a good idea until something will change substantially.

SilverLinings89:
Tried all commands, looked over the changes. Looks flawless to me.

README.md
```diff
@@ -76,7 +76,7 @@ This will compile your project and store the build artifacts in the `dist/` dire
 
 ## Running unit tests
 
-To execute unit tests with the [Karma](https://karma-runner.github.io) test runner, use the following command:
+To execute unit tests with the [Jest](https://jestjs.io/) testing framework, use the following command:
 
 ```bash
 ng test
```
ac-jorellanaf, line 82: This is incorrect no?

ac-jorellanaf (commented):
Looks god, bar one thing

README.md
Cito: Good point, yes, this is a relic from the auto generated documentation. The new section on unit testing is further below. I will add a documentation clean-up task to the next sprint to remove all these auto-generated parts and proof-read the whole documentation.

PR #30, opened 2025-01-10
Add cashew (GSI-1111)

Cito:
This PR adds cachew with the "implicit" strategy which will cache all GET requests in memory by default.

To test, open the browser on the home page, then navigate to the browse page and back. Note that the stats endpoint is only requested once. If you change the strategy to "explicit", then you will see a new request being made when you navigate back.

This PR also changes the class based http interceptor for adding the CSRF header to a functional one since class based interceptors are deprecated.


ac-jorellanaf (approved):
Works perfectly 👍

PR #31, opened 2025-01-13
Do not use staging URL if not needed

Cito:
If running without backend and fake OIDC, set base URL to localhost even if configurated differently.

Only show the hint to open the browser with the staging URL when needed.


ac-jorellanaf (approved):
Everything working great

PR #32, opened 2025-01-13
Adding page titles to the router.

SilverLinings89:
This is just a minor update to provide page titles in the router.

### DIFF
--- a/src/app/app.config.ts
+++ b/src/app/app.config.ts
@@ -11,17 +11,18 @@ import {
 } from '@angular/core';
 import { provideNoopAnimations } from '@angular/platform-browser/animations';
 import { provideAnimationsAsync } from '@angular/platform-browser/animations/async';
-import { provideRouter } from '@angular/router';
+import { provideRouter, TitleStrategy } from '@angular/router';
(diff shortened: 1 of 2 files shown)


Cito (changes requested):
It's nice how simple this now is in Angular.

Instead of just "GHGA" I would say "GHGA Data Portal" and make this a constant so that we can change it more easily:  `${PRE_TITLE}Confirm TOTP`. Or even better, use a custom TitleStrategy that adds the prefix as shown [here](https://angular.dev/guide/routing/common-router-tasks#setting-the-page-title).

SilverLinings89:
Will do. However, there are many people who have a lot of tabs open and not a lot of space for tab titles. That's why I usually keep it as short as possible. The Strategy definitely makes sense, will implement it that way and then we can simply change the constant part in one place.

Cito:
Good point. If we have this centralized, we can also easily switch the order, i.e. add the "GHGA Data Portal" at the end of the title instead of putting it upfront. I think that may be even better.

SilverLinings89:
I have moved the implementation to a Title Strategy.

Cito (approved):

PR #34, opened 2025-01-13
Add guards to auth flow pages (GSI-1206)

Cito:
- adds activation guards to all pages that are part of the auth flow


ac-jorellanaf (commented):
The guards work well, but they take several seconds to redirect (with mocks, but that shouldn't matter). Is this something that could be improved?

ac-jorellanaf (approved):

PR #39, opened 2025-01-21
Updating README and LICENSE. Adding a table of contents tool for the …

SilverLinings89:
…readme. Increase release to 1.0.0. Dependency updates. GSI-1254


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "data-portal",
-  "version": "0.0.0",
+  "version": "1.0.0",
```
Cito, line 3: I would leave it at 0.x until it's released and then directly jump to 2.0 in order to not cause confusion with existing legacy frontend versions when this is tested by the team. But I can change that in my clean-up PR.

Cito (approved):

PR #40, opened 2025-01-21
Add and use notification service

Cito:
- add reusable notification service for different types of notifications
- show notifications (mostly error messages) where needed in the existing code
- add tests
- add loading indicators

### DIFF
--- a/src/app/auth/features/confirm-totp/confirm-totp.component.ts
+++ b/src/app/auth/features/confirm-totp/confirm-totp.component.ts
@@ -15,6 +15,7 @@ import { MatButtonModule } from '@angular/material/button';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
 import { AuthService } from '@app/auth/services/auth.service';
+import { NotificationService } from '@app/shared/services/notification.service';
 
 // TODO: Polish this component
 
@@ -34,6 +35,7 @@ import { AuthService } from '@app/auth/services/auth.service';
   styleUrl: './confirm-totp.component.scss',
 })
 export class ConfirmTotpComponent {
+  #notify = inject(NotificationService);
   #authService = inject(AuthService);
 
   codeControl = new FormControl<string>('', [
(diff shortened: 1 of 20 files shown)


Cito (commented):
- add reusable notification service for different types of notifications
- show notifications (mostly error messages) where needed in the existing code
- TBD: add tests
- TBD: add loading indicators

src/app/auth/features/confirm-totp/confirm-totp.component.ts
```diff
@@ -61,13 +63,11 @@ export class ConfirmTotpComponent {
     if (!code) return;
     const verified = await this.#authService.verifyTotpCode(code);
     if (verified) {
-      console.info('Successfully authenticated.');
-      // add toast message here
+      this.#notify.showError('Successfully authenticated.');
```
SilverLinings89, line 66: Is this intentional to call showError with a success message?

SilverLinings89: Ah I see now, this should be showSuccess, right?

SilverLinings89 (approved):
In general, this looks very good. I remarked on a call that should be showSuccess but that's just a minor point. I have two main questions:
1. Should we give the error messages an ID so users can tell us what error ID they got in case they contact the helpdesk? This could possibly help trace the source of the error better in the future.
2. If we use the logging into the function (console.log, console.warn and console.err into showSuccess, showWarn and showError) we could greatly de-duplicate code.

src/app/auth/features/confirm-totp/confirm-totp.component.ts
Cito: Right, I just fixed it.

Cito:
1. Regarding IDs, since the messages show only for some seconds, that would be difficult to remember. And then they have two pieces of information that they need to read. I think if they roughly remember the error message, we can figure it out.

2. Regarding the console logs, I'm not sure we should keep them. If yes, I agree, we could put these in the notification service.

Cito:
Update: We decided to keep the console logging, and I moved it to the notification service. We can also suppress the logging in production mode if needed, but I am not doing that for now.

PR #41, opened 2025-01-21
Move replaceAll logic from a template to a pipe and adding tests.

SilverLinings89:
In general we should never use function calls in templates. The main reason for this is that angular cannot cache the value in that case (because it cannot automatically determine if the function is pure or not) and therefore, this has to be re-computed on every ui update cycle. 

There are linter rules for this but afaik they can't yet tell the difference between signal() and function() and therefore, they also warn or error for signals (which angular handles properly). 

The solution is to put the logic into a pipe (which is pure) and this gives angular the needed information to compute the value as infrequently as possible.


src/app/shared/utils/underscore_to_space.pipe.spec.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * These are the unit tests for the UnderscoreToSpace pipe.
+ * @copyright The GHGA Authors
+ * @license Apache-2.0
+ */
+
+import { UnderscoreToSpace } from './underscore_to_space.pipe';
+
+describe('DateToYearPipe', () => {
```
ac-jorellanaf, line 9: Wrong name?

ac-jorellanaf, line 9:
```suggestion
describe('UnderscoreToSpacePipe', () => {
```

PR #43, opened 2025-01-22
Polished Dataset Summary

ac-jorellanaf:
Closes GSI-1286


SilverLinings89 (approved):
As a general hint, I didn't want to bloat this too much, but this page may be helpful: https://angular.dev/guide/templates/variables I see that we have lots of repeated logic like summary().experiments_summary.stats.experiment_methods. Similar access patterns repeat. In that case, you can use let to define a new variable and give it a simpler name. For example `@let methods = summary().experiments_summary.stats.experiment_methods`.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.scss
```diff
@@ -1,3 +1,51 @@
...
+      label-text-size: var(--mat-sys-label-small-size),
+    )
+  );
+}
+
+:host ::ng-deep .mat-mdc-standard-chip .mdc-evolution-chip__action--primary {
+  padding: 0 6px;
+}
+
+p {
```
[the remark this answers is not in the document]
Cito: Also, ::ng-deep should be avoided if possible.

src/mocks/data.ts
Cito, line 1: I already changed this in the latest PR, you need to merge main and probably resolve merge conflicts.

src/mocks/data.ts
```diff
@@ -123,7 +113,7 @@ export const datasetSummary = {
   accession: 'GHGAD588887987',
   description:
     'Test dataset for Metadata Repository get dataset details call. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Vel risus commodo viverra maecenas accumsan lacus vel. Sit amet risus nullam eget felis eget nunc lobortis mattis. Iaculis at erat pellentesque adipiscing commodo. Volutpat consequat mauris nunc congue. At lectus urna duis convallis convallis tellus id interdum velit. Gravida cum sociis natoque penatibus et. Mauris in aliquam sem fringilla ut morbi. Ultrices gravida dictum fusce ut. At consectetur lorem donec massa sapien faucibus et molestie.',
-  type: ['Test Type'],
+  type: ['Test Type', 'Test Type 2'],
```
Cito, line 116: I think this is now called types instead of type in the backend. I fixed this but you should double check.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.html
```diff
@@ -1,95 +1,210 @@
...
+                }}
+              </p>
+              @if (summary().samples_summary.stats.tissues.length > 0) {
+                <ul class="ms-6">
+                  @for (
+                    tissue of summary().samples_summary.stats.tissues;
+                    track tissue.value
+                  ) {
+                    <li>
+                      <mat-chip class="mb-0.5 text-xs"
```
Cito, line 106: It's a bit unclear to me why we sometimes use chips and sometimes ordinary bullet points. But it is like that in the legacy app and this is our reference for now. We may need to revisit this later.

Cito: If the four sections would be more uniform (e.g. not one using chips and one bullet points), we could encapsulate them in components to avoid repetition. In the future we should do that more.

src/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <title>GHGA Data Portal</title>
+    <title>The German Human Genome-Phenome Archive</title>
```
Cito, line 5: We're already using "GHGA Data Portal" in the title strategy in Angular since it is shorter and more specific.

You see this title quickly before Angular kicks in, so would be also less flickering in the tab.

Cito:
> I see that we have lots of repeated logic like summary().experiments_summary.stats.experiment_methods

Besides template variables, you can also move logic and application knowledge from the template to the component by defining computed signals in the component that directly return the data to be shown in the template. I did that for the global stats.

Cito (changes requested):
Looks nice.

Some small changes are still needed and the resolution of the merge conflict.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.html
Cito: Anyway, no action needed right now.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.scss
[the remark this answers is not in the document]
ac-jorellanaf: Tailwind do not overwrite material classes, which is a problem and why we need custom CSS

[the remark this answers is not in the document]
Cito: That's true. If there is no overwrite provided by Material, then this is the only way to modify it.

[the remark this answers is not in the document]
ac-jorellanaf: Also yes, prettifying the variables to avoid this sort of repetition is a good idea. I hadn't done it before out of expediency.

src/index.html
ac-jorellanaf: I just changed it to match the current data portal, but I can leave it as it was no problem.

Cito: We decided to use a shorter title because the space on the tabs can be limited.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.scss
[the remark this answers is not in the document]
Cito: My suggestion is to put the following into the styles.css sheet:

```scss
// Chips with less padding
.small-chips {
  .mat-mdc-chip .mat-mdc-chip-action {
    padding: 0 6px;
  }
}
```

Then you can wrap all sections where you want to have smaller-than-usual chips into an element with a "small-chips" class.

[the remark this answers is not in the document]
ac-jorellanaf: This did not work for me. The issue is that this element is in the shadow DOM. We could maybe try to make a class in the global stylesheet but even the angular docs themselves state that ng-deep has to be used for certain libraries. The alternative is we don't try to change this at all and style the component around this.

src/index.html
ac-jorellanaf: What about using `The GHGA Archive`? as a compromise? 😄

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.scss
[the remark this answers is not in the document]
SilverLinings89: I'm sorry, the comment function includes a weird area of the code. I had wanted to comment on the p { ... } which doesn't use ng-deep.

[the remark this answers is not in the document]
Cito: @ac-jorellanaf I tried this yesterday, and it worked for me. So can you try again? Put that snippet in styles.css and add a "small-chips" class to the surrounding div.

The Shadow DOM is by default only emulated. Angular adds random attributes to the elements of the component, and also adds these to the selectors of the styles you define for the component. That's why you can't style the inner elements because they don't have these attributes. But if you put the styles into the global styles.css, the selectors are not modified and can style the inner elements. We are also using this to style the bg color of the snackbar.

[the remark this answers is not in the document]
ac-jorellanaf: Ah sorry. I somehow missed that you wanted to put it in the base stylesheet. Yes I applied it there already and it works fine.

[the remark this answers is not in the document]
Cito: Such custom tweaks to the material components often are useful across the whole application, so it is not too awful to have them in the global styles.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.ts
```diff
@@ -34,8 +34,21 @@ export class DatasetExpansionPanelComponent {
   #metadata = inject(MetadataService);
 
   hit = input.required<Hit>();
+  hitContent = computed(() => this.hit().content);
 
   summary = this.#metadata.datasetSummary;
+  studiesSummary = computed(() => this.summary().studies_summary);
+  studiesSummaryStats = computed(() => this.summary().studies_summary.stats);
+  samplesSummary = computed(() => this.summary().samples_summary);
+  samplesSex = computed(() => this.summary().samples_summary.stats.sex);
```
Cito, line 43:
```suggestion
  samplesSex = computed(() => this.samplesSummaryStats().sex);
```

similarly in other places

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.ts
```diff
@@ -34,8 +34,21 @@ export class DatasetExpansionPanelComponent {
   #metadata = inject(MetadataService);
 
   hit = input.required<Hit>();
+  hitContent = computed(() => this.hit().content);
 
   summary = this.#metadata.datasetSummary;
+  studiesSummary = computed(() => this.summary().studies_summary);
+  studiesSummaryStats = computed(() => this.summary().studies_summary.stats);
```
Cito, line 41:
```suggestion
  studiesSummaryStats = computed(() => this.studiesSummary().stats);
```

src/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <title>The German Human Genome-Phenome Archive</title>
+    <title>GHGA Archive</title>
```
Cito, line 5: The whole product is the "GHGA Archive", but the web client is called "GHGA Data Portal". This also makes it clear that this is not the GHGA homepage. And the "A" in GHGA already stands for "Archive" so "GHGA Archive" is actually an awkward duplication.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.html
```diff
@@ -1,95 +1,183 @@
...
+        <div class="flex grow flex-col gap-y-8">
+          <div class="flex">
+            <div class="me-3 pt-4 text-5xl text-neutral-700">
+              <mat-icon inline>description</mat-icon>
+            </div>
+            <div>
+              <p><strong>File Summary:</strong></p>
+              <p>
+                <strong>{{ studiesSummary().count }}</strong> File{{
+                  studiesSummary().count === 1 ? '' : 's'
```
Cito, line 130: Might be good to put pluralization into a pipe or something. (Must not be done in this PR, just saying).

Cito: See here for example https://prototyp.digital/blog/pluralization-in-angular-ngplural

src/index.html
Cito: That's also why the URL was chosen to be data.ghga.de and not archive.ghga.de.

src/app/metadata/features/dataset-expansion-panel/dataset-expansion-panel.component.html
```diff
@@ -1,95 +1,183 @@
...
+                <strong>{{ experimentsSummary().count }}</strong> Experiment{{
+                  experimentsSummary().count === 1 ? '' : 's'
+                }}
+              </p>
+              <p>
+                <strong
+                  >{{ experimentsSummary().stats.experiment_methods.length }}
+                </strong>
+                Platform{{
+                  experimentsSummary().stats.experiment_methods.length === 1 ? '' : 's'
```
Cito, line 161: It seems you forgot to use some of the custom signals like experimentSummaryStats.

Cito (changes requested):
Looks, but I still have a problem with the title.

src/app/shared/utils/final-s-plural.pipe.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * This pipe takes a js Date object and transforms it to a string of the year. This can be used in the page header but also in data views.
```
Cito, line 2: Adapt comment

src/app/shared/utils/final-s-plural.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @copyright The GHGA Authors
+ * @license Apache-2.0
+ */
+
+import { Pipe, PipeTransform } from '@angular/core';
+
+/**
+ * This pipe can be used to display a year as a string from a Date object.
```
Cito, line 10: Adapt comment

src/app/shared/utils/final-s-plural.pipe.spec.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * These are the unit tests for the DateToYear (example) pipe.
```
Cito, line 2: adapt comment

src/app/shared/utils/final-s-plural.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+@Pipe({
+  name: 'addPluralS',
+})
+export class AddPluralS implements PipeTransform {
+  /**
+   * The transform method executes the business logic of the Pipe
+   * @param count This function will get the number of items and returns an s if it is different than 1 (e.g. 1 method vs. 0 methods)
+   * @returns An s or an empty string, depending on the number of items
+   */
+  transform(count: number): string {
```
Cito, line 21: You could also use the word as input, and the count as an additional parameter that is passed to the pipe.

Usage would then be like `'apple' | addPluralS numApples`

But just converting the number to an s is also ok.

tests/home.spec.ts
```diff
@@ -9,7 +9,7 @@ import { expect, test } from '@playwright/test';
 test('has proper title', async ({ page }) => {
   await page.goto('/');
 
-  await expect(page).toHaveTitle('Home | GHGA Data Portal');
+  await expect(page).toHaveTitle('Home | GHGA Archive');
```
Cito, line 12: This test still needs to be adapted

Cito (changes requested):
Just some small adaptations. You don't need to rewrite the pipe, it's just an idea.

Cito (approved):
Thanks

PR #47, opened 2025-01-24
Filtering and Search in the Browse view

SilverLinings89:
I first implemented search and then noticed that filter facets were closely related so I included it. 

This contains: 
- Search
- Filter Facets
- Chips
- and some tests

### DIFF
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.html
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.html
@@ -2,7 +2,7 @@
   <div class="col-span-9 col-start-4">
     @if (lastSearchQuery() !== '') {
       <mat-chip class="float-left"
-        ><button (click)="clearSearchQuery()">
+        ><button class="align-middle" (click)="clearSearchQuery()">
           <mat-icon>Close</mat-icon></button
         >Keyword: {{ lastSearchQuery() }}</mat-chip
       >
@@ -12,7 +12,10 @@
         @for (option of facet.options; track option.value) {
           @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
             <mat-chip class="float-left">
-              <button (click)="removeFacet(facet.key + '#' + option.value)">
+              <button
+                class="align-middle"
+                (click)="removeFacet(facet.key + '#' + option.value)"
+              >
                 <mat-icon>Close</mat-icon>
               </button>
               {{ facet.name }}: {{ option.value }}
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.html
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.html
@@ -46,24 +46,22 @@
     >
     <mat-chip-set aria-label="search and filters">
       @if (lastSearchQuery() !== '') {
-        <mat-chip class="float-left mr-2"
-          ><button class="align-middle" (click)="clearSearchQuery()">
-            <mat-icon>Close</mat-icon></button
-          >Keyword: {{ lastSearchQuery() }}</mat-chip
+        <mat-chip (removed)="clearSearchQuery()">
+          Keyword: {{ lastSearchQuery() }}
+          <button matChipRemove>
+            <mat-icon>cancel</mat-icon>
+          </button></mat-chip
         >
       }
       @for (facet of searchResults().facets; track facet.key) {
         @if (lastSearchFilterFacets()[facet.key]) {
           @for (option of facet.options; track option.value) {
             @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
-              <mat-chip class="float-left mr-2">
-                <button
-                  class="align-middle"
-                  (click)="removeFacet(facet.key + '#' + option.value)"
-                >
-                  <mat-icon>Close</mat-icon>
-                </button>
+              <mat-chip (removed)="removeFacet(facet.key + '#' + option.value)">
                 {{ facet.name }}: {{ option.value }}
+                <button matChipRemove>
+                  <mat-icon>cancel</mat-icon>
+                </button>
               </mat-chip>
             }
           }
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.html
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.html
@@ -41,34 +41,36 @@
     </div>
   </aside>
   <div class="col-span-9 col-start-4">
-    <mat-chip class="float-right m-1"
-      >Total Datasets: {{ searchResults().count }}</mat-chip
-    >
-    <mat-chip-set aria-label="search and filters">
-      @if (lastSearchQuery() !== '') {
-        <mat-chip (removed)="clearSearchQuery()">
-          Keyword: {{ lastSearchQuery() }}
-          <button matChipRemove>
-            <mat-icon>cancel</mat-icon>
-          </button></mat-chip
-        >
-      }
-      @for (facet of searchResults().facets; track facet.key) {
-        @if (lastSearchFilterFacets()[facet.key]) {
-          @for (option of facet.options; track option.value) {
-            @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
-              <mat-chip (removed)="removeFacet(facet.key + '#' + option.value)">
-                {{ facet.name }}: {{ option.value }}
-                <button matChipRemove>
-                  <mat-icon>cancel</mat-icon>
-                </button>
-              </mat-chip>
+    <div class="inline-block w-full pb-4">
+      <mat-chip class="float-right m-1"
+        >Total Datasets: {{ searchResults().count }}</mat-chip
+      >
+      <mat-chip-set aria-label="search and filters">
+        @if (lastSearchQuery() !== '') {
+          <mat-chip (removed)="clearSearchQuery()">
+            Keyword: {{ lastSearchQuery() }}
+            <button matChipRemove>
+              <mat-icon>cancel</mat-icon>
+            </button></mat-chip
+          >
+        }
+        @for (facet of searchResults().facets; track facet.key) {
+          @if (lastSearchFilterFacets()[facet.key]) {
+            @for (option of facet.options; track option.value) {
+              @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
+                <mat-chip (removed)="removeFacet(facet.key + '#' + option.value)">
+                  {{ facet.name }}: {{ option.value }}
+                  <button matChipRemove>
+                    <mat-icon>cancel</mat-icon>
+                  </button>
+                </mat-chip>
+              }
             }
           }
         }
-      }
-    </mat-chip-set>
-    <div class="mb-4 grid grid-cols-12 bg-primary px-6 py-2 text-white">
+      </mat-chip-set>
+    </div>
+    <div class="mb-4 grid w-full grid-cols-12 bg-primary px-6 py-2 text-white">
       <span class="col-span-2">Dataset ID</span><span class="col-span-10">Title</span>
     </div>
     @if (searchIsLoading()) {
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.html
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.html
@@ -1,7 +1,7 @@
 <div class="grid grid-cols-12 grid-rows-none gap-5">
   <div class="col-span-9 col-start-4">
     @if (lastSearchQuery() !== '') {
-      <mat-chip class="float-left"
+      <mat-chip class="float-left mr-2"
         ><button class="align-middle" (click)="clearSearchQuery()">
           <mat-icon>Close</mat-icon></button
         >Keyword: {{ lastSearchQuery() }}</mat-chip
@@ -11,7 +11,7 @@
       @if (lastSearchFilterFacets()[facet.key]) {
         @for (option of facet.options; track option.value) {
           @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
-            <mat-chip class="float-left">
+            <mat-chip class="float-left mr-2">
               <button
                 class="align-middle"
                 (click)="removeFacet(facet.key + '#' + option.value)"
@@ -28,15 +28,16 @@
   </div>
   <aside class="col-span-3">
     <div class="rounded-lg border p-3">
-      <form (submit)="submit()" #form="ngForm">
+      <form [formGroup]="searchFormGroup" (keydown.enter)="$event.preventDefault()">
         <mat-form-field class="w-full">
           <input
             type="search"
             matInput
             placeholder="Search"
             name="searchTerm"
+            [formControl]="searchFormControl"
             [(ngModel)]="searchTerm"
-            (keyup.enter)="submit()"
+            (keyup.enter)="submit($event)"
           />
         </mat-form-field>
         @for (facet of searchResults().facets; track facet.key) {
@@ -59,8 +60,8 @@
           </div>
         }
         <div class="mt-8 grid grid-cols-2 gap-5">
-          <button mat-stroked-button (click)="clear()">Clear</button>
-          <button mat-flat-button (click)="submit()" type="submit">Filter</button>
+          <button mat-stroked-button (click)="clear($event)">Clear</button>
+          <button mat-flat-button (click)="submit($event)" type="submit">Filter</button>
         </div>
       </form>
     </div>
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.html
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.html
@@ -1,31 +1,4 @@
-<div class="grid grid-cols-12 grid-rows-none gap-5">
-  <div class="col-span-9 col-start-4">
-    @if (lastSearchQuery() !== '') {
-      <mat-chip class="float-left mr-2"
-        ><button class="align-middle" (click)="clearSearchQuery()">
-          <mat-icon>Close</mat-icon></button
-        >Keyword: {{ lastSearchQuery() }}</mat-chip
-      >
-    }
-    @for (facet of searchResults().facets; track facet.key) {
-      @if (lastSearchFilterFacets()[facet.key]) {
-        @for (option of facet.options; track option.value) {
-          @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
-            <mat-chip class="float-left mr-2">
-              <button
-                class="align-middle"
-                (click)="removeFacet(facet.key + '#' + option.value)"
-              >
-                <mat-icon>Close</mat-icon>
-              </button>
-              {{ facet.name }}: {{ option.value }}
-            </mat-chip>
-          }
-        }
-      }
-    }
-    <mat-chip class="float-right">Total Datasets: {{ searchResults().count }}</mat-chip>
-  </div>
+<div class="grid grid-cols-12 grid-rows-none gap-x-5 gap-y-3">
   <aside class="col-span-3">
     <div class="rounded-lg border p-3">
       <form [formGroup]="searchFormGroup" (keydown.enter)="$event.preventDefault()">
@@ -39,6 +12,7 @@
             [(ngModel)]="searchTerm"
             (keyup.enter)="submit($event)"
           />
+          <mat-icon matSuffix>search</mat-icon>
         </mat-form-field>
         @for (facet of searchResults().facets; track facet.key) {
           <div>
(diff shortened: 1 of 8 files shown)


src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -29,32 +32,92 @@ import { DatasetExpansionPanelComponent } from '../dataset-expansion-panel/datas
...
+    );
+  }
 
-  #skip = 0;
+  /**
+   * Sets the parameters for the search and triggers it in the metadataSearch Service
+   */
+  submit(): void {
+    this.updateMetadataServiceSearchTerms();
+    this.#metadataSearch.triggerReload();
```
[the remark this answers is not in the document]
Cito: Similarly in other places below.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -76,5 +135,46 @@ export class MetadataBrowserComponent implements OnInit {
...
+   * This function transforms the checkbox change event to the data required in updateFacet(...) and calls it
+   * @param input the Event from the checkbox
+   */
+  onFacetFilterChanged(input: MatCheckboxChange) {
+    const facetData = input.source.name?.split('#');
+    if (facetData?.length != 2) {
+      return;
+    }
+    const facetKey = facetData[0];
+    const facetOption = facetData[1];
```
Cito, line 148:
```suggestion
    const [facetKey, facetOption] = facetData;
```

src/app/metadata/services/metadataSearch.service.ts
```diff
@@ -27,21 +28,23 @@ export class MetadataSearchService {
   #className = signal<string | undefined>(undefined);
   #limit = signal<number | undefined>(undefined);
   #skip = signal<number | undefined>(undefined);
+  #query = signal<string | undefined>(undefined);
+  #facets = signal<FacetFilterSetting>({});
```
Cito, line 32: Maybe use a single signal "queryParams" that includes all of the above as an object, since we are not setting them individually anyway.

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
...
         <mat-form-field class="w-full">
-          <input type="search" matInput placeholder="Search" />
+          <input
+            type="search"
+            matInput
+            placeholder="Search"
+            name="searchTerm"
+            [(ngModel)]="searchTerm"
+            (keyup.enter)="submit()"
+          />
```
Cito, line 40: This doesn't work for me. Somehow, the input field gets cleared when I press Enter.

Cito (changes requested):
Looks good, see some comments.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -29,32 +32,92 @@ import { DatasetExpansionPanelComponent } from '../dataset-expansion-panel/datas
     MatPaginatorModule,
     MatChipsModule,
     MatFormFieldModule,
+    MatIconModule,
     MatInputModule,
     MatButtonModule,
+    FormsModule,
```
ac-jorellanaf, line 38:
```suggestion
    FormsModule,
    ReactiveFormsModule,
```

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -5,15 +5,18 @@
  */
 
 import { Component, computed, effect, inject, OnInit } from '@angular/core';
-import { FormControl } from '@angular/forms';
+import { FormControl, FormsModule } from '@angular/forms';
```
ac-jorellanaf, line 8:
```suggestion
import { FormControl, FormsModule, ReactiveFormsModule } from '@angular/forms';
```

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
 <div class="grid grid-cols-12 grid-rows-none gap-5">
   <div class="col-span-9 col-start-4">
+    @if (lastSearchQuery() !== '') {
+      <mat-chip class="float-left"
+        ><button class="align-middle" (click)="clearSearchQuery()">
+          <mat-icon>Close</mat-icon></button
+        >Keyword: {{ lastSearchQuery() }}</mat-chip
+      >
```
ac-jorellanaf, line 8:
```suggestion
      <mat-chip (removed)="clearSearchQuery()">
        Keyword: {{ lastSearchQuery() }}
        <button matChipRemove>
          <mat-icon>cancel</mat-icon>
        </button></mat-chip
      >
```

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
...
+          @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
+            <mat-chip class="float-left">
+              <button
+                class="align-middle"
+                (click)="removeFacet(facet.key + '#' + option.value)"
+              >
+                <mat-icon>Close</mat-icon>
+              </button>
+              {{ facet.name }}: {{ option.value }}
+            </mat-chip>
```
ac-jorellanaf, line 22:
```suggestion
            <mat-chip (removed)="removeFacet(facet.key + '#' + option.value)">
              {{ facet.name }}: {{ option.value }}
              <button matChipRemove>
                <mat-icon>cancel</mat-icon>
              </button>
            </mat-chip>
```

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
...
+          }
+        }
+      }
+    }
     <mat-chip class="float-right">Total Datasets: {{ searchResults().count }}</mat-chip>
   </div>
   <aside class="col-span-3">
     <div class="rounded-lg border p-3">
-      <form>
+      <form (submit)="submit()" #form="ngForm">
```
ac-jorellanaf, line 31:
```suggestion
      <form (submit)="submit()">
```
Not sure what this declaration is for?

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
 <div class="grid grid-cols-12 grid-rows-none gap-5">
   <div class="col-span-9 col-start-4">
+    @if (lastSearchQuery() !== '') {
```
ac-jorellanaf, line 3:
```suggestion
    <mat-chip class="float-right m-1"
      >Total Datasets: {{ searchResults().count }}</mat-chip
    >
    <mat-chip-set aria-label="search and filters">
    @if (lastSearchQuery() !== '') {
```

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
...
+              >
+                <mat-icon>Close</mat-icon>
+              </button>
+              {{ facet.name }}: {{ option.value }}
+            </mat-chip>
+          }
+        }
+      }
+    }
     <mat-chip class="float-right">Total Datasets: {{ searchResults().count }}</mat-chip>
```
ac-jorellanaf, line 27:
```suggestion
    </mat-chip-set>
```

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
 <div class="grid grid-cols-12 grid-rows-none gap-5">
```
ac-jorellanaf, line 1:
```suggestion
<div class="grid grid-cols-12 grid-rows-none gap-x-5 gap-y-3">
```
Looks better when the chips are not that separated from the heading of the list

src/app/metadata/features/metadata-browser/metadata-browser.component.html
ac-jorellanaf: Same here

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
ac-jorellanaf: Prevents reloading of the page when submitting the form

src/app/metadata/features/metadata-browser/metadata-browser.component.html
ac-jorellanaf: Some additional CSS magic should be done so that they have a different colour than the total datasets chip.

src/app/metadata/features/metadata-browser/metadata-browser.component.html
ac-jorellanaf: Looks better to use a mat-chip-set when it comes to separation. I also fixed up the floating of the total datasets chip, and added a margin to make it equivalent to the other chips.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
ac-jorellanaf: I agree. It doesn't seem necessary.

src/app/metadata/services/metadataSearch.service.ts
ac-jorellanaf: Well, if we want to implement immediate query search with debounce, we might need it.

ac-jorellanaf (changes requested):
A good implementation, but a few changes required.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
SilverLinings89: I tried to figure this out and my understanding is this: The #globalSummary isn't a signal, it's a resource. Resources are assumed to be heavy workloads. As a consequence, re-processing of the loader has to be triggered manually. Otherwise, even if the signals change, the loader doesn't get called.

src/app/metadata/features/metadata-browser/metadata-browser.component.html
ac-jorellanaf: https://github.com/angular/angular/issues/12643

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
Cito: But the metadata browser does not use the metadata service (which uses the #globalSummary which currently does not have automated loading), but the metadata *search* service. This has automatic loading via the request param which is a signal. That's exactly what "request" is meant to do.

Of course you can also decide to load manually, but then you should remove the request param. However, I find this mechanism pretty elegant. And it still does not load arbitrarily, but only after you call loadQueryParameters because that sets the signals that finally trigger the massQueryUrl signal.

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,28 +1,66 @@
...
-          <input type="search" matInput placeholder="Search" />
+          <input
+            type="search"
+            matInput
+            placeholder="Search"
+            name="searchTerm"
+            [(ngModel)]="searchTerm"
+            (keyup.enter)="submit()"
+          />
         </mat-form-field>
```
ac-jorellanaf, line 41:
```suggestion
        <mat-icon matSuffix>search</mat-icon>
        </mat-form-field>
```

src/app/metadata/services/metadataSearch.service.ts
```diff
@@ -68,8 +81,80 @@ export class MetadataSearchService {
...
+    }
+    massQueryUrl += `class_name=${className}`;
+    for (let facetKey in facets) {
+      for (let option in facets[facetKey]) {
+        massQueryUrl += `&filter_by=${facetKey}&value=${facets[facetKey][option]}`;
+      }
+    }
+    if (query) {
+      massQueryUrl += '&';
+      massQueryUrl += `query=${encodeURIComponent(query)}`;
```
Cito, line 148: why not do it in one line (as you do above), see also below

SilverLinings89: Previously I had some logic to check if it should be ? or & but didn't remove it all when I removed that.

src/app/metadata/features/metadata-browser/metadata-browser.component.html
SilverLinings89: should be fixed now.

src/app/metadata/features/metadata-browser/metadata-browser.component.html
SilverLinings89: nice!

SilverLinings89:
I switched around the order of the content a bit so because previously, if there were lots of chips, the filter and search sidebar was aligned with the results table, making it have whitespace at the top. Now it's always at the top of the left column.

src/app/metadata/services/metadataSearch.service.ts
SilverLinings89: done

src/app/metadata/features/metadata-browser/metadata-browser.component.html
SilverLinings89: done

src/app/metadata/features/metadata-browser/metadata-browser.component.html
SilverLinings89: removed

src/app/metadata/features/metadata-browser/metadata-browser.component.html
SilverLinings89: much better, agreed.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
SilverLinings89: This doesn't seem to work for me. Maybe we can look into it together but for me, the searchResults resource does not trigger a reload if the massQueryUrl gets updated. I think I'm not understanding something here.

src/app/metadata/services/metadataSearch.service.ts
```diff
@@ -27,21 +28,23 @@ export class MetadataSearchService {
...
+  });
 
   #searchResults = resource({
     request: computed(() => ({
-      className: this.#className(),
-      limit: this.#limit(),
-      skip: this.#skip(),
+      url: this.#massQueryUrl(),
     })),
     loader: (param) => {
```
Cito, line 44:
```suggestion
    loader: ({request}) => {
```
I find this a bit nicer than the obscure `param` attribute. And then you can just access request instead of param.request.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
Cito: I double checked and it definitely works without manual reload. But keep in mind that we are using cashew(), so it will not cause a http fetch if you ran the same query already. Also, the query is currently only started when you click on "Filter" or remove a facet.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -29,32 +32,104 @@ import { DatasetExpansionPanelComponent } from '../dataset-expansion-panel/datas
...
+  performSearch(): void {
+    this.updateMetadataServiceSearchTerms();
+    this.#metadataSearch.triggerReload();
+  }
 
-  #skip = 0;
+  /**
+   * Pushes the local filters, search term and page setup to the search service.
+   */
+  updateMetadataServiceSearchTerms(): void {
```
Cito, line 81: this and some other methods could be private

Cito:
I'm also getting a deprecation warning regarding ngModel in the dev console. Maybe you can look into that.

src/app/metadata/services/metadataSearch.service.ts
SilverLinings89: fixed.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
SilverLinings89: I found my problem and it's a little weird: If I start the server with --with-backend I see two datasets and using the search query GHGAD33646518790164 reduces that down to just that set. Even if the explicit reload is disabled. So I see that that works now. However, if I use the staging backend and do a similar search for a dataset ID it doesnt work for me and that was my issue. I will  remove the explicit reload().

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -70,15 +70,15 @@ export class MetadataBrowserComponent implements OnInit {
   /**
    * Syncs the data between this component and the search service and initiates a new search.
    */
-  performSearch(): void {
+  private performSearch(): void {
```
Cito, line 73: Can you use a JS [private property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties) instead? I've started using that new feature everywhere already.

src/app/metadata/services/metadataSearch.service.ts
```diff
@@ -68,8 +81,77 @@ export class MetadataSearchService {
...
+   * The search results loading state as a signal
+   */
+  searchResultsAreLoading: Signal<boolean> = computed(
+    () => this.#searchResults.isLoading() ?? false,
+  );
+
+  /**
+   * This triggers the search Service to reload the search results
+   */
+  triggerReload = () => {
```
Cito, line 94: This method can now actually be removed. Or mention in the comments that id does not need to be called.

Cito (changes requested):
Looks good. Made two small remarks, and there is still a merge conflict.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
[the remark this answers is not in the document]
Cito: Yes, the staging backend is static, it always gives the same replies.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
SilverLinings89: will do.

src/app/metadata/services/metadataSearch.service.ts
SilverLinings89: done

Cito (approved):

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,50 +1,94 @@
...
+                  {{ facet.name }}: {{ option.value }}
+                  <button matChipRemove>
+                    <mat-icon>cancel</mat-icon>
+                  </button>
+                </mat-chip>
+              }
+            }
+          }
+        }
+      </mat-chip-set>
```
ac-jorellanaf, line 70:
```suggestion
    <div class="inline-block grid w-full grid-cols-12 pb-4">
      <mat-chip-set aria-label="search and filters" class="col-span-10">
        @if (lastSearchQuery() !== '') {
          <mat-chip (removed)="clearSearchQuery()">
            Keyword: {{ lastSearchQuery() }}
            <button matChipRemove>
              <mat-icon>cancel</mat-icon>
            </button></mat-chip
          >
        }
        @for (facet of searchResults().facets; track facet.key) {
          @if (lastSearchFilterFacets()[facet.key]) {
            @for (option of facet.options; track option.value) {
              @if (lastSearchFilterFacets()[facet.key].indexOf(option.value) !== -1) {
                <mat-chip (removed)="removeFacet(facet.key + '#' + option.value)">
                  {{ facet.name }}: {{ option.value }}
                  <button matChipRemove>
                    <mat-icon>cancel</mat-icon>
                  </button>
                </mat-chip>
              }
            }
          }
        }
      </mat-chip-set>
      <div class="col-span-2 col-start-11">
        <mat-chip class="float-right m-1"
          >Total Datasets: {{ searchResults().count }}</mat-chip
        >
      </div>
```
Gets rid of annoying whitespace changes

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,50 +1,94 @@
-<div class="grid grid-cols-12 grid-rows-none gap-5">
-  <div class="col-span-9 col-start-4">
-    <mat-chip class="float-right">Total Datasets: {{ searchResults().count }}</mat-chip>
-  </div>
+<div class="grid grid-cols-12 grid-rows-none gap-x-5 gap-y-3">
   <aside class="col-span-3">
```
ac-jorellanaf, line 2: Not my favourite look change, since now the search and the list header are not aligned at the start.

ac-jorellanaf (changes requested):
Last few changes

src/app/metadata/features/metadata-browser/metadata-browser.component.html
SilverLinings89: I tried with several filters on and it also looked really weird when they wrap ...

SilverLinings89: Is there a solution to fix both?

ac-jorellanaf: In the current data portal it's as it was before, where the seachbar remains aligned with the header regardless. I am not sure which looks good.

SilverLinings89: OK, I'll revert it to the old state.

src/app/metadata/services/metadataSearch.service.ts
SilverLinings89: Is it OK if I keep this as it is and merge and we figure that out when we decide about immediate search?

PR #50, opened 2025-01-27
Add work package service

Cito:
This PR adds the Angular service with the two methods needed to interact with the work package service on the backend:
- find all datasets for which work packages can be created
- create a work package

The first one is usable via three corresponding signals, but I kept the return value of the second one as an observable, so you need to subscribe. A resource does not seem to be the right paradigm here, and converting to a promise or signal doesn't look very appealing either. So I suggest we stick to observables for mutations.

The PR also shows how we can simplify the code using an rxResource, and how to test it. This is a bit tricky - you need to use the new and hardly known flushEffects if the service is not running in the context of a component. Also, waiting for a tick was needed to get the results from the loader.

### DIFF
--- a/.devcontainer/data-portal.yaml
+++ b/.devcontainer/data-portal.yaml
@@ -14,6 +14,7 @@ base_url: http://127.0.0.1:8080
 auth_url: /api/auth
 mass_url: /api/mass
 metldata_url: /api/metldata
+wps_url: /api/wps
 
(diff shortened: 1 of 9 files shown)


ac-jorellanaf (approved):
There is no component to run a live test, but the service tests and the code all look good.

SilverLinings89 (approved):

PR #51, opened 2025-01-28
Route parameters for the browse view

SilverLinings89:
Covers search, filter facets, page size and skip.

### DIFF
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.ts
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.ts
@@ -14,11 +14,15 @@ import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatIconModule } from '@angular/material/icon';
 import { MatInputModule } from '@angular/material/input';
 import { MatPaginatorModule, PageEvent } from '@angular/material/paginator';
+import { ActivatedRoute, Router } from '@angular/router';
 import { FacetFilterSetting } from '@app/metadata/models/facet-filter';
 import { MetadataSearchService } from '@app/metadata/services/metadataSearch.service';
 import { FacetActivityPipe } from '@app/metadata/utils/facetActivity.pipe';
 import { NotificationService } from '@app/shared/services/notification.service';
 import { DatasetExpansionPanelComponent } from '../dataset-expansion-panel/dataset-expansion-panel.component';
+
+const DEFAULT_PAGE_SIZE = 10;
+const DEFAULT_SKIP_VALUE = 0;
 
 /**
  * This is the metadata browser component
@@ -44,12 +48,14 @@ import { DatasetExpansionPanelComponent } from '../dataset-expansion-panel/datas
 export class MetadataBrowserComponent implements OnInit {
   #notify = inject(NotificationService);
   #metadataSearch = inject(MetadataSearchService);
+  #route = inject(ActivatedRoute);
+  #router = inject(Router);
   #className = 'EmbeddedDataset';
-  #skip = 0;
+  #skip = DEFAULT_SKIP_VALUE;
   #pageEvent!: PageEvent;
 
   facetData: FacetFilterSetting = {};
-  pageSize = 10;
+  pageSize = DEFAULT_PAGE_SIZE;
   searchFormControl = new FormControl('');
   searchFormGroup = new FormGroup({
     searchTerm: this.searchFormControl,
@@ -65,20 +71,40 @@ export class MetadataBrowserComponent implements OnInit {
    * On init, define the default values of the search variables
    */
   ngOnInit(): void {
-    this.#updateMetadataServiceSearchTerms();
+    const { pageSize, search, filters, skip } =
+      this.#router.routerState.snapshot.root.queryParams;
+    if (pageSize) {
+      this.pageSize = parseInt(pageSize);
+    } else {
+      this.pageSize = DEFAULT_PAGE_SIZE;
+    }
+    if (search) {
+      this.searchTerm = search;
+    }
+    if (filters) {
+      const paramVals = JSON.parse(decodeURIComponent(filters));
+      if (paramVals) {
+        this.facetData = paramVals;
+      }
+    }
+    if (skip) {
+      this.#skip = parseInt(skip);
+    }
+
+    this.#updateMetadataServiceSearchTermsAndRoute();
   }
 
   /**
    * Syncs the data between this component and the search service and initiates a new search.
    */
   #performSearch(): void {
-    this.#updateMetadataServiceSearchTerms();
+    this.#updateMetadataServiceSearchTermsAndRoute();
   }
 
   /**
    * Pushes the local filters, search term and page setup to the search service.
    */
-  #updateMetadataServiceSearchTerms(): void {
+  #updateMetadataServiceSearchTermsAndRoute(): void {
     this.#metadataSearch.loadQueryParameters(
       this.#className,
       this.pageSize,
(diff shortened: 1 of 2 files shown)


src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -65,27 +71,59 @@ export class MetadataBrowserComponent implements OnInit {
...
       this.facetData,
     );
+    this.#router.navigate([], {
+      relativeTo: this.#route,
+      queryParams: {
+        skip: this.#skip !== DEFAULT_SKIP_VALUE ? this.#skip : undefined,
+        search: this.searchTerm !== '' ? this.searchTerm : undefined,
+        filters:
+          Object.keys(this.facetData).length !== 0
+            ? encodeURIComponent(JSON.stringify(this.facetData))
```
ac-jorellanaf, line 122: I would prefer if we avoided using these heavily-encoded JSON-style objects, especially since they make URLs unwieldy and difficult to read, and keep it consistent with the backend query URL (also by using `q=` instead of `search=`)

ac-jorellanaf (changes requested):

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
SilverLinings89: I can do that, here's my reasoning why I didn't so far: I think it's an antipattern to rely on the order of the parameters. For example: If I call ?key=abc&val=def&key=abc&val=ghi then the value of routerState.snapshot.root.queryParams will contain arrays (because both key and val have multiple values). They will be key: ['abc', 'abc'] and val: ['def', 'ghi']. Now we make the assumption, that this is in order but I'm not sure we actually can. Query parameters (to my knowledge) are assumed to have no order so it could be, for example in the Angular URL parser that the order is not maintained but instead something is ordered alphabetically. A bookmarking tool or browser could also do that to check if two urls are the same. The version I implemented circumvents this issue but I agree it doesn't lead to pretty URLs (which is why I at least checked if the value is set to keep the URL as short as possible).

SilverLinings89: We control this behavior on the backend but I'm not sure this is safe in the frontend...

Cito: Interesting discussion. Yes, sadly, the order of params is generally not guaranteed.

On the backend, we rely on urllib.parse.parse_qsl and starlette's QueryParams preserving order (but neither of them is actually documented to guarantee this).

I saw that you reverted to using the old f and q parameters on the frontend which I think is the best solution for now and even backward compatible.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -65,27 +71,59 @@ export class MetadataBrowserComponent implements OnInit {
...
+      queryParams: {
+        skip: this.#skip !== DEFAULT_SKIP_VALUE ? this.#skip : undefined,
+        q: this.searchTerm !== '' ? this.searchTerm : undefined,
+        f:
+          Object.keys(this.facetData).length !== 0
+            ? encodeURIComponent(this.#facetDataToString(this.facetData))
+            : undefined,
+        pageSize: this.pageSize !== DEFAULT_PAGE_SIZE ? this.pageSize : undefined,
+      },
+    });
```
Cito, line 126: See above. Would be more elegant to only change the route and then use the same method as on init to load the new query params.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -161,6 +199,55 @@ export class MetadataBrowserComponent implements OnInit {
...
+    const ret: FacetFilterSetting = {};
+    for (const facet of filterStrings) {
+      if (facet.indexOf(':') !== -1) {
+        const [facetName, facetValue] = facet.split(':');
+        if (ret[facetName]) {
+          ret[facetName].push(facetValue);
+        } else {
+          ret[facetName] = [facetValue];
+        }
+      }
```
Cito, line 245:
```suggestion
      const [facetName, facetValue] = facet.split(':' ,2);
      if (facetValue !== undefined) {
        if (ret[facetName]) {
          ret[facetName].push(facetValue);
        } else {
          ret[facetName] = [facetValue];
        }
      }
```

the 2 makes sure it also works with values that contain colons

Cito (changes requested):
Looks good, but I suggest some changes regarding param names and control flow.

Also, the console.logs should be removed before merging. We should use console.log only for temporary logging during development. If the logging should be permanent, we can use log.debug or log.info with proper messages.

SilverLinings89:
I added all the suggestions and made the url the source of truth. Sorry for the logs that I forgot to remove - my bad.

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -65,20 +71,27 @@ export class MetadataBrowserComponent implements OnInit {
...
-   * Pushes the local filters, search term and page setup to the search service.
+   * This function takes the current URL, determines the query parameters and sets them accordingly in the metadata search service.
    */
-  #updateMetadataServiceSearchTerms(): void {
+  #loadSearchTermsFromRoute(): void {
+    const { s, q, f, p } = this.#router.routerState.snapshot.root.queryParams;
+    this.pageSize = parseInt(p) || DEFAULT_PAGE_SIZE;
+    if (q) {
+      this.searchTerm = q;
+    }
```
Cito, line 85:
```suggestion
    this.searchTerm = q || '';
```

this allows resetting the search term if it was set previously

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -65,20 +71,27 @@ export class MetadataBrowserComponent implements OnInit {
...
+    if (f) {
+      const paramVals = this.#facetDataFromString(decodeURIComponent(f));
+      if (paramVals) {
+        this.facetData = paramVals;
+      }
+    }
+    if (s) {
+      this.#skip = parseInt(s);
+    }
```
Cito, line 94:
```suggestion
    this.#skip = parseInt(s) || DEFAULT_SKIP_VALUE;
```

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -161,6 +200,50 @@ export class MetadataBrowserComponent implements OnInit {
...
+    if (Object.keys(facetData).length === 0) {
+      return '';
+    }
+    let facetStrings = [];
+    for (const key in facetData) {
+      for (const index in facetData[key]) {
+        facetStrings.push(key + ':' + facetData[key][index]);
+      }
+    }
+    return facetStrings.join(';');
```
Cito, line 219: You could reduce here and below, but not sure if that would be really easier to read.

Cito (changes requested):
Looks good, but check if it is possible to reset query params (see comments below).

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
SilverLinings89: personally I prefer this version...

SilverLinings89: In reduce, we would need to individually handle either the first or last case. Join only adds the ';' in-between the items.

Cito (approved):

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
Cito: You could just build the array with reduce, and then still join. But you're right, reduce does not make it more readable.

PR #54, opened 2025-02-01
Improve dataset details page (GSI-1323)

Cito:
- Change route from browse/id to /dataset/id
- Show dataset ID in the title
- Add table sorting and pagination
- Improve unit tests
- Activate and adapt corresponding e2e test
- Various minor improvements

### DIFF
--- a/src/app/metadata/features/dataset-details-page/dataset-details-page.component.spec.ts
+++ b/src/app/metadata/features/dataset-details-page/dataset-details-page.component.spec.ts
@@ -22,6 +22,8 @@ class MockMetadataService {
   loadDatasetDetails = () => undefined;
 }
 
+import { screen } from '@testing-library/angular';
+
 /**
  * Mock the dataset information service as needed for the dataset details
  */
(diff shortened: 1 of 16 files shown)


src/app/metadata/features/dataset-details-page/dataset-details-page.component.ts
```diff
@@ -113,33 +143,103 @@ export class DatasetDetailsPageComponent implements OnInit {
...
+  filesColumns: string[] = COLUMNS.files.split(' ');
+
+  #experimentsSortingDataAccessor = (experiment: Experiment, key: string) => {
+    switch (key) {
+      case 'method':
+        return experiment.experiment_method.name;
+      case 'platform':
+        return experiment.experiment_method.instrument_model;
+      default:
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
SilverLinings89, line 157: This seems problematic. This can return an ExperimentMethod which wouldn't be compatible with the type required by the uses (which expect string | number)

src/app/metadata/features/dataset-details-page/dataset-details-page.component.ts
```diff
@@ -113,33 +143,103 @@ export class DatasetDetailsPageComponent implements OnInit {
...
+      case 'status':
+        return sample.case_control_status;
+      case 'sex':
+        return sample.individual.sex;
+      case 'phenotype':
+        return sample.individual.phenotypic_features_terms.join(', ');
+      case 'tissue':
+        return sample.biospecimen_tissue_term;
+      default:
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
SilverLinings89, line 173: See above.

src/app/metadata/features/dataset-details-page/dataset-details-page.component.ts
```diff
@@ -113,33 +143,103 @@ export class DatasetDetailsPageComponent implements OnInit {
...
+      case 'origin':
+        return file.file_category;
+      case 'size':
+        return file.file_information?.size;
+      case 'location':
+        return file.file_information?.storage_alias;
+      case 'hash':
+        return file.file_information?.sha256_hash;
+      default:
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
SilverLinings89, line 191: same here.

src/app/shared/utils/parse-bytes.pipe.spec.ts
```diff
@@ -12,15 +12,26 @@ describe('ParseBytesPipe', () => {
...
+  });
+
+  it('should return 25 KB for 25,000 Bytes', () => {
+    const pipe = new ParseBytes();
+    const result = pipe.transform(25_000);
     expect(result).toBe('25\u00A0kB');
   });
 
-  it('should return 2.5 MB for 25000000 Bytes', () => {
+  it('should return 2.5 MB for 2,500,000 Bytes', () => {
```
SilverLinings89, line 32: Do we have a rule somewhere to use decimal over binary multipliers aka. KB=1000B rather than 1024?

SilverLinings89 (changes requested):
Minor Comments, maybe we should discuss the forced anys. Other than that LGTM.

src/app/metadata/features/dataset-details-page/dataset-details-page.component.spec.ts
```diff
@@ -19,16 +19,18 @@ import { DatasetDetailsPageComponent } from './dataset-details-page.component';
 class MockMetadataService {
```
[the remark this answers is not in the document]
Cito: Good idea. One reason is that we usually need to only mock a small part of the functionality for the respective test, and the mock may provide behavior that is (only) useful for that test.

We could also use the real services and just mock the underlying http client, using our existing mock data or the MSW. But you're right there may be also other testing strategies besides mocks.

src/app/metadata/features/dataset-details-page/dataset-details-page.component.ts
Cito: Improved a bit in 1d13aec.

src/app/shared/utils/parse-bytes.pipe.spec.ts
Cito: Changed in 7516617 to correspond to how Windows file explorer does it (even though it's not correct).

PR #55, opened 2025-02-03
Feature/basal account page gsi 1296

SilverLinings89:
As per issue 1296. Layout different than in old ui. May need some work still.


src/app/app.routes.ts
```diff
@@ -46,6 +46,15 @@ export const routes: Routes = [
...
     title: 'Download or Upload Datasets',
   },
+  {
+    path: 'profile',
+    canActivate: [() => inject(AuthService).guardAuthenticated()],
+    loadComponent: () =>
+      import('./portal/features/account/account.component').then(
+        (m) => m.AccountComponent,
+      ),
+    title: 'User Profile',
```
Cito, line 56:
```suggestion
    title: 'User Account',
```

We originally called it the "Profile" page, but then changed it to "Account" page, we should this wording now everywhere.

src/app/app.routes.ts
```diff
@@ -46,6 +46,15 @@ export const routes: Routes = [
       ),
     title: 'Download or Upload Datasets',
   },
+  {
+    path: 'profile',
```
Cito, line 50:
```suggestion
    path: 'account',
```

src/app/portal/features/account/account.component.html
```diff
@@ -0,0 +1,81 @@
...
+  </mat-card-header>
+  <mat-card-content class="my-4 text-base">
+    <p class="text-base">
+      We will communicate with you via this email address:
+      <span class="font-bold">{{ auth.email() }}</span>
+    </p>
+    <p>
+      You can change this email address in your
+      <a
+        class="underline"
```
Cito, line 22: I added consistent styling for links in the main section of the page in the global style sheet in #56, so that we don't need to add a color or "underline" any more if we want to use the standard link style.

src/app/portal/features/account/account.component.html
```diff
@@ -0,0 +1,81 @@
+<h1 class="text-center">
+  <span class="text-3xl">User Account</span><br />
+  <span class="text-3xl">{{ auth.fullName() }}</span>
+</h1>
+
```
Cito, line 5: This should also show whether you're logged in as a data steward somewhere. (In the legacy app this was shown in the green message at the top.)

src/app/portal/features/account/account.component.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @license Apache-2.0
+ */
+
+import { Component, inject } from '@angular/core';
+import { MatCardModule } from '@angular/material/card';
+import { MatIconModule } from '@angular/material/icon';
+import { AuthService } from '@app/auth/services/auth.service';
+
+/**
+ * This Component shows data about the current user and allows for modification of the profile (via child components).
```
Cito, line 13:
```suggestion
 * This Component shows data about the current user and allows managing their IVAs.
```

Cito (approved):
This page would look a bit less boring with some contrasting colors and icons.

E.g. the heading could have a larger "user profile" icon and the individual sections smaller icons on the top left corner.

We can also experiment with giving the card headers an orange background with white text or something like that.

But we can also do this in an additional "polish account page" task.

src/app/portal/features/account/account.component.html
```diff
@@ -0,0 +1,81 @@
+<h1 class="text-center">
+  <span class="text-3xl">User Account</span><br />
+  <span class="text-3xl">{{ auth.fullName() }}</span>
+</h1>
```
Cito, line 4: Maybe would be better to use two headings of different levels or to use different colors for the title and the name.

Cito (dismissed):

src/app/portal/features/account/account.component.html
SilverLinings89: I'll remove it here and then it will be fixed one #56 is merged

Cito (approved):

PR #61, opened 2025-02-10
Login dropdown (GSI-1327)

Cito:
- Add dropdown menu instead of the login/logout button with immediate action.
- Handle all possible authentication states.
- Show initials of the user's name when authenticated.
- Make LS Login account management URL configurable.


src/app/portal/features/account-button/account-button.component.ts
```diff
@@ -0,0 +1,93 @@
...
+import { MatIconModule } from '@angular/material/icon';
+import { MatMenuModule } from '@angular/material/menu';
+import { MatTooltipModule } from '@angular/material/tooltip';
+import { Router } from '@angular/router';
+import { AuthService } from '@app/auth/services/auth.service';
+import { ConfigService } from '@app/shared/services/config.service';
+import { getInitials } from '@app/shared/utils/initials';
+
+/**
+ * Account button component
```
SilverLinings89, line 19: I'm struggling a bit with these description (also in the components I contribute. If the description is basically just the name, then I think there is no reason for the documentation. However, I also can't really think of a rule that would guarantee a better result.

src/app/portal/features/account/account.component.html
```diff
@@ -3,9 +3,9 @@ <h1 class="text-center">
 </h1>
 <h2>
   <span class="align-middle text-2xl"
-    >{{ auth.fullName() }}
-    @if (auth.role() === 'data_steward') {
-      <mat-chip> Data Steward </mat-chip>
+    >{{ fullName() }}
+    @if (roleName()) {
```
SilverLinings89, line 7: I wasn't sure here. Do we have a definite, finished list of roles? Because it felt to me like a user could have multiple roles (Like for example "user", "data steward" and "admin"). But this depends on the explicit set of roles we have to determine if there should just be one ("data stewards" are always also "users" etc.) or if this is more like individual permissions a user has or doesn't have. If the latter holds then the user could have multiple roles. This is just food for thought though.

src/app/shared/utils/initials.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it('does not change case if names are already capitalized', () => {
+    expect(getInitials('Alice Brown')).toBe('AB');
+  });
+
+  it('works with all kinds of whitespace', () => {
+    expect(getInitials('alice\u00a0brown')).toBe('AB');
+    expect(getInitials('alice\tbrown')).toBe('AB');
+    expect(getInitials('alice \t \u00a0 \r \n brown')).toBe('AB');
```
SilverLinings89, line 47: Perfect coverage :) well done!

src/app/shared/utils/initials.ts
SilverLinings89, line 1: Somehow I have the feeling this should be a pipe since we only really use it in a template (we compute it in the component but we don't use the transformed version in logic anywhere) and therefore this somehow feels like a pipe to me, but I also can't really tell if there would be any advantage. It's fine for me either way.

SilverLinings89 (dismissed):
Some minor remarks but could also be merged like this.

src/app/portal/features/account-button/account-button.component.ts
Cito: It also feels a bit awkward to me. Also, most files only contain one class but we require a file and a class description, and then it's unclear where to put the long description. We also don't want to repeat it in both places. And the component names aer often pretty self-explanatory anyway.

OTOH, since we don't require documentation for class members and arrow functions, it is not enforced that e.g. (computed) signals are documented, even though that may be often not self-evident.

So we may need to rethink our documentation approach a bit.

src/app/portal/features/account/account.component.html
Cito: The role concept is not yet finalized. Currently we support only one role, data steward. But in the future we may also have admins. No role means "ordinary user". It's still unclear whether we need multiple roles. If the admin role includes the data steward role, we would not need that, but we may want to separate them. In that case, the backend also needs to be adapted, since it currently does no support multiple roles. Also, whether we want to have roles that apply only to a certain data hub.

SilverLinings89 (approved):

PR #62, opened 2025-02-11
Add IVA section to account page (GSI-1297)

Cito:
WIP. Missing: Button to add new IVA, tests.


src/app/metadata/models/dataset-information.ts
```diff
@@ -11,9 +11,10 @@ export interface DatasetInformation {
 
 export interface FileInformation {
   accession: string;
-  size: number;
-  sha256_hash: string;
-  storage_alias: string;
+  // the remaining fields may be undefined
```
ac-jorellanaf, line 14: This doesn't seem sound at all, but I assume this is an issue with the backend (and is unrelated to this PR)?

src/app/shared/services/confirmation.service.spec.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * Short module description
```
ac-jorellanaf, line 2:
```suggestion
 * Tests for shared confirmation dialog service
```

src/app/shared/ui/confirm-dialog/confirm-dialog.component.ts
```diff
@@ -8,8 +8,11 @@ import { Component, inject } from '@angular/core';
 import { MatButtonModule } from '@angular/material/button';
 import {
   MAT_DIALOG_DATA,
+  MatDialogActions,
```
ac-jorellanaf, line 11: Are these dialogs really necessary? Are they not available by just importing the MatDialogModule?
It's a minor thing anyways.

src/app/verification-addresses/services/iva.service.ts
```diff
@@ -128,11 +186,27 @@ export class IvaService {
...
+        map(({ id }) => id),
+        tap((id) => this.#addIvaLocally({ id, type, value, userId })),
+      );
+  }
+
+  /**
+   * Delete an IVA locally
+   * @param ivaId - the ID of the IVA to delete
+   */
+  #deleteIvaLocally(ivaId: string): void {
```
ac-jorellanaf, line 199: Can't we have a function to do local addition/state modification of IVAs as well? Or just use the update also for deletion?

src/app/verification-addresses/services/iva.service.ts
```diff
@@ -146,31 +220,53 @@ export class IvaService {
       userId = this.#auth.user()?.id;
       if (!userId) return throwError(() => new Error('Not authenticated'));
     }
-    return this.#http.delete<null>(`${this.#userIvasUrl(userId)}/${ivaId}`);
+    return this.#http
+      .delete<null>(`${this.#userIvasUrl(userId)}/${ivaId}`)
+      .pipe(tap(() => this.#deleteIvaLocally(ivaId)));
```
ac-jorellanaf, line 225: What if there is an error when deleting? Would the pipe not run?

ac-jorellanaf (approved):
Everything works to the letter. I only have a few minor comments.

src/app/metadata/models/dataset-information.ts
Cito: Yes, the backend sometimes doesn't return this fields in the response. This is not a bug, intended behavior and defined in the API. By marking the fields as undefined here, we get a TypeScript warning when using them unshielded, so we don't create the issues we had in the legacy app.

src/app/verification-addresses/services/iva.service.ts
Cito: Yes, the `tap()` taps only into the happy path. You can also handle the error path in tap(), but handling only the happy case is exactly what we want.

ac-jorellanaf (dismissed):

src/app/metadata/models/dataset-information.ts
Cito: Added explanation in comment now

src/app/shared/ui/confirm-dialog/confirm-dialog.component.ts
Cito: Do you mean the imports here or generally having a specialized confirmation dialog?

I think the imports are necessary because we use the  mat-dialog-actions directives etc. And I added the specialized confirmation dialog because we need this kind of dialog frequently, so having a shared component is handy and also ensures a consistent appearance.

src/app/verification-addresses/services/iva.service.ts
Cito: We have that already, in #addIvaLocally and #updateIvaStateLocally. The latter is shared by the different state modification methods.

ac-jorellanaf (approved):

src/app/shared/ui/confirm-dialog/confirm-dialog.component.ts
ac-jorellanaf: I meant the imports yes sorry.

src/app/verification-addresses/services/iva.service.ts
ac-jorellanaf: Ah no I meant having a single update method instead of all these separate ones that do similar things.

Cito: Ok. But I think they are distinct enough to warrant their own methods. Also, I think it's good practice to have smaller functions that do only one thing (aka Curly's Law or Single Responsibility Principle):
- https://activeprogrammerhq.com/functions-clean-code/

PR #64, opened 2025-02-14
Fix work-package token creation issues (GSI-1356)

Cito:
- Improve error messages in dev launcher
- Fix issue when global stats do not return detailed stats
- Add helper function for deriving a user friendly error from the backend response
- Improve error handling when work-package creation fails
- This has been tested against staging to make sure the expected error is displayed to the user

### DIFF
--- a/src/app/work-packages/features/work-package/work-package.component.ts
+++ b/src/app/work-packages/features/work-package/work-package.component.ts
@@ -79,34 +79,69 @@ export class WorkPackageComponent {
   }
 
   /**
+   * Remove headers and fix padding of the given base64 encoded key
+   * @param key - the key to be trimmed
+   * @returns the trimmed key
+   */
+  #trimKey(key: string): string {
+    // allow and trim headers and footers for public keys
+    key = key
+      .replace(/-----(BEGIN|END) CRYPT4GH PUBLIC KEY-----/, '')
+      .replace(/^[\s-]+/, '')
+      .replace(/[\s-=]+$/, '');
+    const fix = key.length % 4;
+    if (fix) {
+      key += '='.repeat(4 - fix);
+    }
+    return key;
+  }
+
+  /**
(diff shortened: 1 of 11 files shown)


src/app/work-packages/features/work-package/work-package.component.ts
```diff
@@ -77,68 +78,121 @@ export class WorkPackageComponent {
...
+  /**
+   * Check whether the given key is a valid public key.
+   * @param key - the key in question
+   * @returns an internal error code
+   */
+  #checkPubKey(key: string): number {
+    key = key.trim();
+    if (!key) {
+      return 1; // key is empty
+    }
```
SilverLinings89, line 108:
```suggestion
    if (!key) {
      return 1; // key is empty
    }
    key = key.trim();
```
I think the other case is captured in line 114 but the argument might be undefined and in that case, key.trim() would fail I think.

SilverLinings89 (dismissed):

src/app/work-packages/features/work-package/work-package.component.ts
Cito: The key parameter is guaranteed to be a string. But it could contain only whitespace.

Cito: I have removed the first check for emptiness now since it's already covered by the second one.

SilverLinings89 (approved):

src/app/work-packages/features/work-package/work-package.component.ts
Cito: If we cannot control the input and/or it is really critical then you shouldn't rely on the typescript checks only, that's true. In this case it's a private method and called with the value of the signal which is a string because it is initialized as an empty string and bound to the value of the input. And the worst thing that could happen is that it throws an error.

PR #69, opened 2025-02-21
Add a filter component to the IVA-Manager (GSI-1362)

Cito:


src/app/verification-addresses/features/iva-manager-filter/iva-manager-filter.component.html
```diff
@@ -1 +1,90 @@
...
+            (click)="this.name.set('')"
+          >
+            <mat-icon>cancel</mat-icon>
+          </button>
+        </div>
+        <div class="flex gap-2">
+          <mat-form-field class="grow">
+            <mat-label>Changed dates ranging from</mat-label>
+            <input matInput [matDatepicker]="fromPicker" [(ngModel)]="fromDate" />
+            <mat-hint>MM/DD/YYYY</mat-hint>
```
SilverLinings89, line 27: This is the same format I used in the other component. I was wondering: There are some stylistic choices (Date format, AE vs BE, capitalization etc.) we make. Should we document them somewhere?

src/app/verification-addresses/features/iva-manager-filter/iva-manager-filter.component.scss
```diff
@@ -0,0 +1,5 @@
+.large-icon {
+  font-size: 48px;
+  width: 48px;
+  height: 48px;
```
SilverLinings89, line 4: tailwind: size-12

src/app/verification-addresses/services/iva.service.ts
```diff
@@ -109,7 +136,28 @@ export class IvaService {
...
+      if (filter.toDate) {
+        const toDate = filter.toDate.toISOString();
+        ivas = ivas.filter((iva) => iva.changed <= toDate);
+      }
+      if (filter.state) {
+        ivas = ivas.filter((iva) => iva.state === filter.state);
+      }
+    }
+    return ivas;
+  });
```
SilverLinings89, line 160: This will have somewhat untypical behavior if there's a space in the search (because it wouldn't search for the individual words but for an exact match of the full term including the space) or is that not possible here?

SilverLinings89 (approved):
Again, just minor stuff.

src/app/verification-addresses/features/iva-manager-filter/iva-manager-filter.component.scss
Cito: Will check again, but sometimes the specificity of the tailwind classes was not sufficient.

src/app/verification-addresses/services/iva.service.ts
Cito: Not sure if we want/need that here. It's only relevant for the name. If you type "John Doe" you usually want to find John Doe, not people whose last name is John or Doe.

src/app/verification-addresses/features/iva-manager-filter/iva-manager-filter.component.html
Cito: Let's commit it like it is now and then create a PR to adapt it everywhere.

src/app/verification-addresses/features/iva-manager-filter/iva-manager-filter.component.scss
Cito: Checked it, the problem is that the width of the mat-icon is more specific. You can only overwrite it in the style sheet or in an inline style.

ac-jorellanaf (approved):
Works fine but date formatting should be changed to something non-US even if not ISO (it confused me trying to test already)

Cito:
Switching to ISO as date format everywhere is now a separate task on the board.

SilverLinings89 (approved):

PR #71, opened 2025-02-25
Feature/access request refactorings gsi 1433

SilverLinings89:


src/app/access-requests/services/data-access.service.ts
```diff
@@ -124,9 +128,30 @@ export class DataAccessService {
     () => this.#accessRequests.value() ?? [],
   );
   grantedAccessRequests = computed(() =>
-    this.accessRequests().filter((ar: AccessRequest) => ar.status == 'allowed'),
+    this.accessRequests()
+      .filter((ar: AccessRequest) => ar.status == 'allowed')
+      .map((request: AccessRequest) => {
+        const expiryDate = new Date(request.access_ends);
```
Cito, line 134: Btw, we also have an `access_starts` date. I'm not sure if we ever would set this to a future date. but maybe we should check it here as well to be on the safe side. Instead of `isExpired` you wold then use an `isValid` flag that combines both checks. Or just use `daysRemaining` and set it to 0 or null if it's invalid.

Cito (approved):
See comments below.

It would help when reviewig if you add a short description what features the PR is intended to cover and which parts will be implemented later.

src/app/access-requests/services/data-access.service.ts
SilverLinings89: I'll do this in a future PR.

Cito (approved):

PR #73, opened 2025-02-25
Added hash abbreviation pipe (GSI-1428)

ac-jorellanaf:
Added hash abbreviation pipe
Used abbr tag to display full hash, but this is only barebones.
A better option should be explored
Improved some documentation for the existing pipes


SilverLinings89 (dismissed):
LGTM

src/app/shared/utils/short-hash.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+  name: 'shortHash',
+})
+export class ShortHash implements PipeTransform {
+  /**
+   * The transform method executes the business logic of the Pipe
+   * @param hash The hash to shorten
+   * @returns The shortened string
+   */
+  transform(hash: string): string {
+    if (hash.length < 7) return hash;
```
Cito, line 22: If the has hash hash has 7 chars, 1234567, then it will become 123456... which is even longer...

I would expect that the output of such a pipe is always shorter than the input.

Cito: Also, would be good to make the number 7 configurable as a size parameter with default value 7 (you can pass additional params to  pipes),

Cito: GitHub abbreviates commit hashes with 7 chars, maybe we should also use 7 by default (instead of 6).

src/app/shared/utils/short-hash.pipe.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { Pipe, PipeTransform } from '@angular/core';
+
+/**
+ * This pipe is used to shorten a hash or other string to its first seven characters plus an ellipsis at the end.
+ */
+@Pipe({
+  name: 'shortHash',
+})
+export class ShortHash implements PipeTransform {
```
Cito, line 15: I suggest to rename this to something like "Truncate" because it can be generally useful for any text output, not just for hash values.

src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -467,7 +467,9 @@ <h4 class="inline-flex items-center">
                   File hash
                 </th>
                 <td mat-cell *matCellDef="let file">
-                  {{ file.file_information?.sha256_hash }}
+                  <abbr title="{{ file.file_information?.sha256_hash }}">{{
+                    file.file_information?.sha256_hash | shortHash
+                  }}</abbr>
```
Cito, line 472: Would be nice to have an option to copy the full sha. GitHub offers this for the commit sha in several places, using a small "copy" button. Alternatively, a right-click menu option would also do. Should not be too difficult to implement. Could then be an ordinary span.

Cito (changes requested):

src/app/shared/utils/short-hash.pipe.ts
ac-jorellanaf: It is 7 characters, not 6! Seems like a classic off-by-one error 😄 
But I will fix it so it does the ellipsis only for strings longer than 10 characters so it never returns longer strings than the original.

Cito: Yes, you can add a size parameter with a default of 10 that would guarantee to always output max 10 chars (including the dots). Including the dots, you would have 7 digits from the hash (just like in GitHub).

Aletrnatively, and maybe better, use a single unicode ellipsis char instead of the three dots. Then the default size would be 8.

src/app/shared/utils/truncate-string.pipe.ts
```diff
@@ -0,0 +1,31 @@
...
+   * The transform method executes the business logic of the Pipe
+   * @param str The string to shorten
+   * @param output_max_length The number of characters from the original string to return (excluding ellipses)
+   * @param input_max_length The minimum length of string from which to start shortening
+   * @returns The shortened string
+   */
+  transform(
+    str: string,
+    output_max_length: number = 7,
+    input_max_length: number = 10,
```
Cito, line 26: I'm not sure whether the 10 now makes sense now because we don't have 3 dots any more, but just one ellipsis char. And I think one param would be sufficient, you can assume that input-max-length is output-max-length + 1. And the param should use camel case or just be called "size".

Cito (approved):
Nice, works like I imagined. Just some small suggestions.

src/app/shared/utils/truncate-string.pipe.ts
ac-jorellanaf: Fixed

src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -467,7 +467,20 @@ <h4 class="inline-flex items-center">
...
+                      class="inline-block max-w-[8ex] overflow-hidden text-ellipsis whitespace-nowrap"
+                      title="{{ file.file_information?.sha256_hash }}"
+                      >{{ file.file_information?.sha256_hash }}</span
+                    >
+                    <button
+                      mat-icon-button
+                      [cdkCopyToClipboard]="file.file_information?.sha256_hash"
+                      (click)="notifyCopied()"
+                    >
+                      <mat-icon>content_copy</mat-icon>
```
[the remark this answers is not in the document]
ac-jorellanaf: It was such a pain to try to make the icon a bit smaller, since I also tried to reduce the button size as well, which was nigh impossible cause angular material does not have the mixins for icon buttons to reduce sizes. Anyways this is done too

src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -462,7 +462,20 @@ <h4 class="inline-flex items-center">
                   File hash
                 </th>
                 <td mat-cell *matCellDef="let file">
-                  {{ file.file_information?.sha256_hash }}
+                  @if (file.file_information?.sha256_hash) {
+                    <span
+                      class="inline-block max-w-[8ex] overflow-hidden text-ellipsis whitespace-nowrap"
```
Cito, line 467:
```suggestion
                      class="inline-block max-w-[10ex] overflow-hidden text-ellipsis whitespace-nowrap"
```

This more or less reliably shows 7 significant digits (more or less because the font is not monospaced, but I think it's good enough).

src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -462,7 +462,20 @@ <h4 class="inline-flex items-center">
...
+                      title="{{ file.file_information?.sha256_hash }}"
+                      >{{ file.file_information?.sha256_hash }}</span
+                    >
+                    <button
+                      mat-icon-button
+                      [cdkCopyToClipboard]="file.file_information?.sha256_hash"
+                      (click)="notifyCopied()"
+                    >
+                      <mat-icon class="text-[20px]/[24px]">content_copy</mat-icon>
```
Cito, line 476: Perfect.

Cito (approved):
This looks very nice now. I just would make the clipped with a bit larger, so that you can really see 7 chars (it was only between 5 or 6 for me).

Cito (approved):
Using a monospaced font is also a good solution.

PR #74, opened 2025-02-26
Access request manager: Filter (GSI-1359)

Cito:
Implement the filter component for the access request manager.

### DIFF
--- a/src/app/access-requests/features/access-request-manager-filter/access-request-manager-filter.component.html
+++ b/src/app/access-requests/features/access-request-manager-filter/access-request-manager-filter.component.html
@@ -36,7 +36,7 @@
         </div>
         <div class="flex gap-2">
           <mat-form-field class="grow">
-            <mat-label>Requested created from</mat-label>
+            <mat-label>Request created from</mat-label>
             <input matInput [matDatepicker]="fromPicker" [(ngModel)]="fromDate" />
             <mat-hint>MM/DD/YYYY</mat-hint>
             <mat-datepicker-toggle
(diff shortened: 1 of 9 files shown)


ac-jorellanaf (dismissed):
All works well, barring some comments:

Should we do partial matching for Dataset IDs too?

Maybe a change of terminology would be useful. Date of request submission vs. requested date? It makes it less ambiguous with the start date of the request.

Cito:
> Should we do partial matching for Dataset IDs too?

If you copy & paste the ID, this is not needed, but if you type it might be useful.

> Maybe a change of terminology would be useful. Date of request submission vs. requested date? It makes it less ambiguous with the start date of the request.

Changed the label in the filter a bit. Should also not be too long, particularly as a table header.

src/app/access-requests/features/access-request-manager-filter/access-request-manager-filter.component.html
```diff
@@ -0,0 +1,104 @@
...
+            aria-label="Remove filter for user name"
+            [class.invisible]="!this.name()"
+            (click)="this.name.set('')"
+          >
+            <mat-icon>cancel</mat-icon>
+          </button>
+        </div>
+        <div class="flex gap-2">
+          <mat-form-field class="grow">
+            <mat-label>Requested created from</mat-label>
```
ac-jorellanaf, line 39:
```suggestion
            <mat-label>Request created from</mat-label>
```

ac-jorellanaf (dismissed):
Also the header change to 'Request created' maybe?

ac-jorellanaf (approved):

PR #75, opened 2025-02-26
Add Dynamic Access Request button GSI 1413

SilverLinings89:
This PR includes a change to the way the create access request dialog is handled. It is no longer controlled by the access request service but instead owned by a button component. This button can be used in multiple places and removes a lot of work from the data access (soon access request) service.


src/app/access-requests/features/dynamic-access-request-button/dynamic-access-request-button.component.ts
```diff
@@ -0,0 +1,107 @@
...
+  status = computed(() => {
+    let newStatus = 'ok';
+    for (let ar of this.#pendingAccessRequests()) {
+      newStatus = ar.status;
+    }
+    for (let gar of this.#activeGrantedAccessRequests()) {
+      newStatus = gar.request.status;
+    }
+    return newStatus;
+  });
```
Cito, line 62:
```suggestion
  #pendingAccessRequests = computed(() =>
    this.#accessRequestService
      .pendingUserAccessRequests()
      .some((ar: AccessRequest) => ar.dataset_id == this.datasetID()),
  );
  #activeGrantedAccessRequests = computed(() =>
    this.#accessRequestService
      .grantedUserAccessRequests()
      .some((grantedAccessRequest: GrantedAccessRequest) => {
        const hasSameId = grantedAccessRequest.request.dataset_id == this.datasetID();
        const isValid = grantedAccessRequest.daysRemaining > 0;
        return hasSameId && isValid;
      }),
  );
  status = computed<AccessRequestStatus>(() => {
    if (this.#activeGrantedAccessRequests())
      return AccessRequestStatus.allowed;
    if (this.#pendingAccessRequests())
      return AccessRequestStatus.pending;
    return AccessRequestStatus.denied;
  });
```

Actually we only need to know whether there are some acess requests in the corresponding state, so this feels a bit more natural than overwriting the state. Unrequested could be treated like denied.

src/app/access-requests/features/dynamic-access-request-button/dynamic-access-request-button.component.html
```diff
@@ -0,0 +1,17 @@
+@switch (status()) {
+  @case ('pending') {
+    <button mat-raised-button [disabled]="true">
+      <mat-icon>key</mat-icon> Pending Access Request ...
+    </button>
+  }
+  @case ('allowed') {
+    <button mat-raised-button routerLink="/work-package">
+      <mat-icon>key</mat-icon>Access the Data
```
Cito, line 9: Or maybe "Download the Dataset"?

src/app/access-requests/features/dynamic-access-request-button/dynamic-access-request-button.component.ts
```diff
@@ -0,0 +1,107 @@
...
+      fromDate: undefined,
+      untilDate: undefined,
+      userId: this.#userId() ?? '',
+    };
+
+    this.#dialogRef = this.#dialog.open(AccessRequestDialogComponent, {
+      data,
+    });
+
+    this.#dialogRef.afterClosed().subscribe((componentData) => {
```
Cito, line 83:
```suggestion
    this.#dialog.open(AccessRequestDialogComponent, {
      data,
    }).afterClosed().subscribe((componentData) => {
```

You don't need to store the dialog ref in the component. You can then remove the declaration as well.

Cito (approved):
Looks good, just some minor suggestions.

The form validation also needs to be done in another PR.

src/app/access-requests/services/access-request.service.ts
```diff
@@ -41,48 +36,7 @@ export class AccessRequestService {
...
-  #performAccessRequest = (data: AccessRequestDialogData) => {
-    data.fromDate?.setUTCHours(0);
-    data.fromDate?.setUTCMinutes(0);
-    data.fromDate?.setUTCSeconds(0);
-    data.fromDate?.setUTCMilliseconds(0);
-    data.untilDate?.setHours(23);
-    data.untilDate?.setMinutes(59);
-    data.untilDate?.setSeconds(59);
-    data.untilDate?.setMilliseconds(999);
+  performAccessRequest = (data: AccessRequestDialogData) => {
```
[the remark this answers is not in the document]
SilverLinings89: Doesn't that break responsibility boundaries a bit?

ac-jorellanaf (changes requested):

src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -4,12 +4,7 @@
       <button mat-icon-button (click)="goBack()">
         <mat-icon>keyboard_return</mat-icon>
       </button>
-      <button
-        mat-stroked-button
-        (click)="accessService.showNewAccessRequestDialog(datasetDetails().accession)"
-      >
-        <mat-icon>key</mat-icon>Request Access
-      </button>
+      <app-dynamic-access-request-button datasetID="datasetDetails().accession" />
```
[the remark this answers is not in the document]
SilverLinings89: good spot, thank you.

src/app/access-requests/features/dynamic-access-request-button/dynamic-access-request-button.component.html
```diff
@@ -0,0 +1,17 @@
+@switch (status()) {
+  @case ('pending') {
+    <button mat-raised-button [disabled]="true">
+      <mat-icon>key</mat-icon> Pending Access Request ...
+    </button>
+  }
+  @case ('allowed') {
+    <button mat-raised-button routerLink="/work-package">
+      <mat-icon>key</mat-icon>Download the Data
```
ac-jorellanaf, line 9:
```suggestion
      <mat-icon>key</mat-icon>Download Data
```
Using 'the Data' sounds wrong, but Download Data is also not perfect.

src/app/access-requests/features/dynamic-access-request-button/dynamic-access-request-button.component.html
```diff
@@ -0,0 +1,17 @@
+@switch (status()) {
+  @case ('pending') {
+    <button mat-raised-button [disabled]="true">
+      <mat-icon>key</mat-icon> Pending Access Request ...
```
ac-jorellanaf, line 4:
```suggestion
      <mat-icon>key</mat-icon> Pending Access Request
```

ac-jorellanaf: or:
```suggestion
      <mat-icon>key</mat-icon> Pending Access Request …
```

src/app/access-requests/features/dynamic-access-request-button/dynamic-access-request-button.component.html
SilverLinings89: "Download Date" gives me 10x more google search results than "Download Data" so I would leave it at this for now.

src/app/access-requests/services/access-request.service.ts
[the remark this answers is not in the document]
Cito: The service may contain logic e.g. using pipes, but I don't think the service should output error messages, their exact wording should be left to the feature component. Also, the component should control the life cycle and further processing of the subscription. For instance, the subscription would be canceled, when the component is destroyed. If the subscription happens in the service, it will continue to run. The pattern also works well with async pipes.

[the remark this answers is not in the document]
Cito: See for instance here: https://angular.dev/guide/http/making-requests#best-practices

Cito (dismissed):

PR #77, opened 2025-02-26
Consistent date format (GSI-1424)

ac-jorellanaf:
I am unsure about the date format in the account page for the access requests etc. It looks less cluttered with a more textual (i.e. with a spelt out month) format, but that would be inconsistent.


Cito (commented):
I think the format with spelled out (or abbreviated) month is also not bad in some cases. Maybe as a rule we can use ISO when we need to spare space (like in tabular form) but spell it out when writing prose and there is enough space?

For the ISO conversion it would be useful to have an isoDate pipe that would be the same as the normal data pipe, but with the iso format as default for the format.

src/app/access-requests/features/access-request-dialog/access-request-dialog.component.html
```diff
@@ -37,7 +37,7 @@ <h2 mat-dialog-title>
         [max]="maxUntilDate"
         [(ngModel)]="untilDate"
       />
-      <mat-hint>MM/DD/YYYY</mat-hint>
+      <mat-hint>DD/MM/YYYY</mat-hint>
```
Cito, line 40: So the output would be ISO and the input European style? Shouldn't we also use ISO for input?

Cito (commented):
Please note that the access request manager (which still has two unmerged PRs) also has several date inputs and outputs which should be covered by this PR as well after merging.

src/app/shared/utils/iso-date.pipe.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    const result = pipe.transform(date);
+    expect(result).toBe('2025-05-11');
+  });
+
+  it('should return 2025-05-12 for the string 12 May 2025', () => {
+    const date = '12 May 2025';
+    const pipe = new isoDatePipe();
+    const result = pipe.transform(date);
+    expect(result).toBe('2025-05-12');
+  });
```
SilverLinings89, line 27: With tests for dates it is typically a good idea to use a day > 12 - that way, mixups in formats like YYYY-MM-DD vs. YYYY-DD-MM can be avoided because usually, months cannot be set to a value greater than 12.

src/app/shared/utils/iso-date.pipe.ts
```diff
@@ -10,17 +10,19 @@ import { Pipe, PipeTransform } from '@angular/core';
...
  */
 @Pipe({
-  name: 'dateToYear',
+  name: 'isoDate',
 })
-export class DateToYearPipe implements PipeTransform {
+export class isoDatePipe implements PipeTransform {
   /**
    * The transform method executes the business logic of the Pipe
    * @param date This function will get the year for the provided date and return it as a string
```
SilverLinings89, line 18: The comments are outdated.

SilverLinings89 (changes requested):
The comments in the old dateToYear pipe file are outdated. Other than that, LGTM

src/app/access-requests/features/access-request-dialog/access-request-dialog.component.html
Cito: I'm not sure about that. On a German keyboard, hyphen and dot are easier to type than slash. Also either DD.MM.YYYY or YYYY-MM-DD (ISO) is the preferred format in official EU documents.

Cito (approved):

Cito:
Merging this, but will create a follow-up PR to centralize the default date format even more.

PR #78, opened 2025-02-27
Add form validation for access request form GSI-1456

SilverLinings89:
This PR includes form validation for the Access Request Dialog. It checks the date bounds, email and requires a description. Otherwise, the submit button is disabled.


ac-jorellanaf (approved):
All works well. In the old data portal though, the user could not change the requested date. I am not sure if this was intended to change in the new one.

PR #79, opened 2025-02-28
Centralize date format (GSI-1424)

Cito:
- Move all date format definitions used in the application to shared/utils/date-formats
- Use these formats in app.config.ts to configure the DateFnsAdapter and the standard DatePipe
- Use the default date format for tabular output, and shortDate format for dates that are part of prose
- Show the defined input date format as hint for date picker input fields

### DIFF
--- a/src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
+++ b/src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
@@ -19,10 +19,10 @@
         <td>{{ data.request_text }}</td>
       </tr>
     </table>
-    <p>The request has been made on {{ data.request_created | date }}.</p>
+    <p>The request has been made on {{ data.request_created | date: 'mediumDate' }}.</p>
     <p>
-      Acess has been requested from {{ data.access_starts | date }} until
-      {{ data.access_ends | date }}.
+      Access has been requested from {{ data.access_starts | date: 'mediumDate' }} until
+      {{ data.access_ends | date: 'mediumDate' }}.
     </p>
     <p>
       Access request is currently <strong>{{ data.status }}</strong
--- a/src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
+++ b/src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
@@ -19,10 +19,14 @@
         <td>{{ data.request_text }}</td>
       </tr>
     </table>
-    <p>The request has been made on {{ data.request_created | date: 'mediumDate' }}.</p>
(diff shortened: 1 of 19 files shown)


SilverLinings89 (dismissed):

src/app/shared/utils/date-formats.ts
```diff
@@ -0,0 +1,35 @@
...
+ * @license Apache-2.0
+ */
+
+import { MatDateFormats } from '@angular/material/core';
+import { enGB } from 'date-fns/locale';
+
+export const DEFAULT_DATE_LOCALE = enGB;
+
+export const DEFAULT_DATE_OUTPUT_FORMAT = 'yyyy-MM-dd';
+export const DEFAULT_DATE_INPUT_FORMAT = 'yyyy-MM-dd';
```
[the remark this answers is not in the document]
Cito: Good point. Made the hint configurable separately from the format now.

src/app/shared/utils/date-formats.ts
```diff
@@ -11,6 +11,7 @@ export const DEFAULT_DATE_LOCALE = enGB;
 
 export const DEFAULT_DATE_OUTPUT_FORMAT = 'yyyy-MM-dd';
 export const DEFAULT_DATE_INPUT_FORMAT = 'yyyy-MM-dd';
+export const DATE_INPUT_FORMAT_HINT = 'YYYY-MM-DD';
```
ac-jorellanaf, line 14: Wouldn't this be repeating code unnecessarily? We could just convert the format used to uppercase

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
```diff
@@ -19,10 +19,10 @@ <h2 mat-dialog-title>Access Request Detail</h2>
         <td>{{ data.request_text }}</td>
       </tr>
     </table>
-    <p>The request has been made on {{ data.request_created | date }}.</p>
+    <p>The request has been made on {{ data.request_created | date: 'mediumDate' }}.</p>
```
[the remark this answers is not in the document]
ac-jorellanaf: Well, it looks more easy to read for me exactly because it's not so regular, so it breaks the monotony of the text

src/app/shared/utils/date-formats.ts
Cito: The idea was to make it more flexible, e.g. the hint could also be "Please enter data as YYYY-MM-DD".

Cito: Also makes the configured values more obvious/visible.

ac-jorellanaf (approved):

PR #80, opened 2025-03-01
Polish account page (GSI-1345)

Cito:
This PR has been created using GitHub Copilot edit in agent mode as an experiment.

### DIFF
--- a/src/app/portal/features/account/account.component.html
+++ b/src/app/portal/features/account/account.component.html
@@ -1,78 +1,130 @@
-<h1 class="text-center">
-  <span class="text-3xl">User Account</span>
-</h1>
-<h2>
-  <span class="align-middle text-2xl"
-    >{{ fullName() }}
-    @if (roleName()) {
-      <mat-chip> {{ roleName() }} </mat-chip>
-    }
-  </span>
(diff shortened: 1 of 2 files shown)


SilverLinings89:
:]

ac-jorellanaf:
This is definitely... something.
It is frankly quite ugly and busy, not to hurt GitHub's feelings.

I would rather this (especially being a designing task) be done manually.

Cito:
And this is already after I have done a lot of back and forth. The first attempt looked a hundred times worse.

The main problem here is that the Agent cannot directly examine the output that it generates in the browser. Therefore, it does not notice if some icon is clipped or badly aligned, you always need to tell it as feedback. And then it makes another attempt, changing some CSS, but it would still be the same (very much like a human).

The other problem is that it over-engineers, adding many redundant style definitions and wrapper elements to solve problems that could be solved in simpler ways. Therefore the code becomes very bloated, looking a bit like Word documents converted to HTML. (Again, it was worse, I already removed some redundant styles.)

Should we create a separate PR or try to manually improve this one?

Cito:
This has been superseded by #85 now.

PR #82, opened 2025-03-03
Implemented the see more, see less badges feature (GSI-1427)

ac-jorellanaf:
- Implemented the see more, see less badges feature
- Moved them to their own component
- Standardised all summary lists into list of badges Fixed some code with the files summary


src/app/metadata/features/dataset-summary/dataset-summary.badges.component.html
```diff
@@ -0,0 +1,29 @@
...
+    let count = $count;
+    let last = $last
+  ) {
+    <li>
+      <mat-chip class="small-chips mb-0.5 text-xs"
+        ><span class="uppercase">{{ item.value | underscoreToSpace }} </span>
+        <strong>{{ item.count }}</strong></mat-chip
+      >
+    </li>
+    @if (count > 3) {
```
Cito, line 15: Please consider the edge cases. As far as I see, if you have exactly 4 items, you will see the button, but it will do nothing except changing its label.

Cito: Would be nice if the button not just says "see more", but "see N more" where N is the number of hidden items. Or if it just says "see all" since it doesn't work gradually anyway.

Cito (changes requested):
Looks good to me, but see comments below.

src/app/metadata/features/dataset-summary/dataset-summary.badges.component.html
ac-jorellanaf: Thanks. It was a classic off-by-one.

src/app/shared/ui/summary-badges/summary-badges.component.spec.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Test the confirm dialog component
```
Cito, line 2:
```suggestion
 * Test the summary badges component
```

Cito (dismissed):

PR #83, opened 2025-03-04
Update to Angular 19.2 (GSI-1426)

Cito:
- Update pnpm to v10
- Update Angular to v19.2
- Update some other more dependencies

TBD:
- make use of the newer Angular features like httpResource and resource defaultValue


ac-jorellanaf (approved):
All works well

PR #87, opened 2025-03-12
Feature/umami integration

SilverLinings89:
I have added tracker events to all clickables with some version of a click handler (href, routerLink, (click), etc.). The idea behind this is to really see what our users are interacting with and what not so we can make informed decisions on where we need to make changes and so we can identify if users understand the UI the way we do (if for example they click "cancel" a lot somewhere, we likely five them a decision too early).
I have can provide the config data for the other stages in slack. Those values aren't secret since we have to include them in the static build of the website and they are intended to be public.

### DIFF
--- a/.devcontainer/data-portal.yaml
+++ b/.devcontainer/data-portal.yaml
@@ -22,3 +22,6 @@ oidc_client_id: ghga-dev-client
 oidc_authority_url: https://login.aai.lifescience-ri.eu/oidc/
 
 ribbon_text: 'Development v$v'
+
+umami_url: https://umami.foundational.testing.ghga.dev/umami
+umami_website_id: 9b8be8ec-ff79-4dda-9af4-f695787e0610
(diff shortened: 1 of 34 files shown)


Cito (approved):
Looks good.

Some minor suggestions below.

Questions:
- Does this also track the (client-side) page navigation out of the box?
- Can we use this to track the e.g. 10 most-used search terms?

SilverLinings89:
> Looks good.
> 
> Some minor suggestions below.
> 
> Questions:
> 
> * Does this also track the (client-side) page navigation out of the box?
The sessions view reports which routes have been visited how often so we can see which URLs have been called.
> * Can we use this to track the e.g. 10 most-used search terms?
There is the option to throw custom events from js so this would be possible to implement. Events can be tracked from js and they can have additional data (see here: https://umami.is/docs/tracker-functions )

Cito (approved):

PR #90, opened 2025-03-12
Improve the Umami integration

Cito:
This PR addresses the following issues:

1. allow disabling Umami via configuration
2. make Umami work with client side navigation
3. mock Umami with MSW (just log what is tracked)

To solve 2. I used the following approach:
- disable the auto-tracking on the script tag (this disables pageview tracking, but sadly also event tracking)
- manually track the client side navigation
- manually track events similar to how Umami would do it with auto-tracking

### DIFF
--- a/src/app/shared/services/umami.service.ts
+++ b/src/app/shared/services/umami.service.ts
@@ -32,6 +32,7 @@ export class UmamiService {
       return;
     }
     const script = document.createElement('script');
+    script.async = true;
     script.src = this.#server_url;
     script.setAttribute('data-website-id', this.#website_id);
     document.head.appendChild(script);
--- a/src/app/shared/services/umami.service.ts
+++ b/src/app/shared/services/umami.service.ts
@@ -12,7 +12,8 @@ import { ConfigService } from './config.service';
 declare global {
   interface Window {
     umami?: {
-      track: (payload: object) => void;
+      track(payload: object): void;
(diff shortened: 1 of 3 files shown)


src/app/shared/services/umami.service.ts
```diff
@@ -27,9 +39,61 @@ export class UmamiService {
...
+          title:
+            this.#router.routerState.snapshot.root.firstChild?.routeConfig?.title ||
+            document.title,
+        });
+      });
+  }
+
+  /**
+   * This method tracks events manually with Umami by subscribing to click events.
+   * We need to do this manually since we switched of auto-tracking in the script tag.
```
SilverLinings89, line 78:
```suggestion
   * We need to do this manually since we switched off auto-tracking in the script tag.
```

SilverLinings89 (approved):
This is a really good solution. Obviously sad that we have to do this that way instead of using auto-tracking but this is an elegant solution that keeps the implementation compatible with auto-tracking in the future if they fix that behavior.

src/app/shared/services/umami.service.ts
SilverLinings89: I will investigate this further tomorrow and merge. I'll also talk to Camill about the config for staging and production.

SilverLinings89:
Checked and the results also look good in the umami Dashboard.

SilverLinings89 (approved):

PR #91, opened 2025-03-13
Polishing of admin pages (GSI-1430)

ac-jorellanaf:
Some additional polishes across various pages.
Some minor improvements to colours and mocks.
Huge PR


src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
```diff
@@ -28,30 +28,36 @@ <h2 mat-dialog-title>Access Request Detail</h2>
...
               <td>
-                <mat-radio-button class="example-radio-button" [value]="iva.id">
+                <mat-radio-button [value]="iva.id">
                   {{ typeAndValue(iva) }}
                 </mat-radio-button>
               </td>
-              <td class="iva{{ iva.state }}">
-                {{ iva.state }}
+              <td class="{{ statusTextClass(iva) }} w-40">
+                {{ statusName(iva) }}
```
SilverLinings89, line 60: Calling functions (other than signals) in a template should be avoided. We don't have a linter rule for it, yet, because those I'm aware of, cannot distinguish between signal calls, which are fine, and function calls, which are not. The problem is that this gets called every time the ui rerenders (which can be very often). For functions, Angular has no way of knowing if the output will change over time, which is different for pipes (which are pure, so no side-effects) and signals (which actively track all side-effects). For functions, however, no such tracking is possible so they have to be re-run every time.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
```diff
@@ -119,10 +129,10 @@ export class AccessRequestManagerDialogComponent implements OnInit {
    * Allow the access request and close the dialog.
```
SilverLinings89, line 129: Just a minor style thing (and I know this was wrong before this commit) - doesn't need fixing right now, but: The name says allow and the comment says allow and close the dialog. Those are different things. Therefore, the method should have a different name or the logic should be split into two distinct parts, because now, the method does stuff that isn't apparent from the name. The same below for deny.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
```diff
@@ -132,26 +142,24 @@ export class AccessRequestManagerDialogComponent implements OnInit {
...
-   * @returns a string that combines the type and value of the IVA
+   * Pre-select the IVA used for an existing access request.
    */
-  typeAndValue(iva: Iva): string {
-    let text = IvaTypePrintable[iva.type];
-    if (iva.value) {
-      text += `: ${iva.value}`;
+  #preSelectIva(): void {
+    if (this.data.iva_id) {
+      this.accessRequestIvaId.set(this.data.iva_id);
```
SilverLinings89, line 158: Similar problem to the allow() deny() thing above.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
```diff
@@ -180,6 +188,46 @@ export class AccessRequestManagerDialogComponent implements OnInit {
...
+  }
+
+  /**
+   * Get a suitable CSS class for an IVA state's status
+   * @param iva - the IVA in question
+   * @returns the class corresponding to the state
+   */
+  statusTextClass(iva: Iva): string {
+    return IVA_STATUS_CLASS[iva.state];
+  }
```
SilverLinings89, line 223: This needs to either be pre-computed and stored as a value or put into a signal. See comment in the template file.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
```diff
@@ -180,6 +188,46 @@ export class AccessRequestManagerDialogComponent implements OnInit {
...
+
+  /**
+   * Get a suitable CSS class for an access request's status
+   * @param access_request - the access request in question
+   * @returns the class corresponding to the state
+   */
+  accessRequestStatusTextClass(access_request: AccessRequest): string {
+    return ACCESS_REQUEST_STATUS_CLASS[access_request.status];
   }
 }
```
SilverLinings89, line 233: See above

src/app/access-requests/features/access-request-manager-list/access-request-manager-list.component.html
```diff
@@ -10,44 +10,51 @@
...
         {{ ar.request_created | date }}
       </td>
     </ng-container>
     <ng-container matColumnDef="status">
-      <th mat-header-cell *matHeaderCellDef class="w-40" mat-sort-header>Status</th>
-      <td mat-cell class="status-{{ ar.status }}" *matCellDef="let ar">
+      <th mat-header-cell *matHeaderCellDef class="min-w-40 capitalize" mat-sort-header>
+        Status
+      </th>
+      <td mat-cell class="{{ statusTextClass(ar) }} capitalize" *matCellDef="let ar">
```
SilverLinings89, line 46: no functions in templates

src/app/shared/utils/capitalise.pipe.spec.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * These are the unit tests for the AddPluralS pipe.
```
SilverLinings89, line 2: copy pasta

src/app/shared/utils/capitalise.pipe.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+    const pipe = new Capitalise();
+    const result = pipe.transform('hello world');
+    expect(result).toBe('Hello world');
+  });
+
+  it('should return 24 for 24', () => {
+    const pipe = new Capitalise();
+    const result = pipe.transform('24');
+    expect(result).toBe('24');
+  });
```
SilverLinings89, line 25: a test for an empty string would also make sense

src/app/verification-addresses/features/code-creation-dialog/code-creation-dialog.component.html
```diff
@@ -1,54 +1,65 @@
...
+        <th class="pe-1 pl-10 text-right">Name:</th>
+        <td>{{ iva.user_name }}</td>
+      </tr>
+      <tr>
+        <th class="pe-1 text-right">E-Mail:</th>
+        <td>{{ iva.user_email }}</td>
+      </tr>
+      <tr>
+        <th class="pe-1 text-right">Via:</th>
+        <td>{{ typeName(iva) }}</td>
```
SilverLinings89, line 16: functions in templates

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
```diff
@@ -1,54 +1,46 @@
 @if (ivasAreLoading()) {
-  <p class="text-base">Loading contact addresses..</p>
+  <p>Loading contact addresses..</p>
```
SilverLinings89, line 2: I've seen you remove the text-base class in many places. From what I can tell it mainly has a style reset role (font-size: var(--text-base); line-height: var(--text-base--line-height);) Is there a reason to remove it?

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
```diff
@@ -1,54 +1,46 @@
...
+      <th mat-header-cell *matHeaderCellDef mat-sort-header>Last modified</th>
+      <td mat-cell *matCellDef="let iva">
         {{ iva.changed | date }}
       </td>
     </ng-container>
     <ng-container matColumnDef="state">
-      <th mat-header-cell *matHeaderCellDef class="w-40" mat-sort-header>Status</th>
-      <td mat-cell class="state{{ iva.state }}" *matCellDef="let iva">
+      <th mat-header-cell *matHeaderCellDef mat-sort-header>Status</th>
+      <td mat-cell class="{{ statusTextClass(iva) }}" *matCellDef="let iva">
```
SilverLinings89, line 33: function in template

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
```diff
@@ -64,18 +56,18 @@
...
             data-umami-event="IVA Manager List Confirm Transmission Clicked"
           >
-            <mat-icon>mark_chat_read</mat-icon>
+            <mat-icon>mark_email_read</mat-icon>
             Confirm transmission
           </button>
         }
         @if (iva.state !== 'Unverified') {
           <button
             (click)="invalidateWhenConfirmed(iva)"
```
SilverLinings89, line 69: Just to make sure: Function calls in event handlers are obviously fine, because they only get called when the event occurs, so no issues here. They are only problematic for  pure output like {{ somefunction(asdf) }} or in class="{{ notHere(asdf)}}" or similar places like [disabled] or style.

SilverLinings89 (changes requested):
Overall, lots of good work. One mayor issue that requires some more changes, is the point of function calls in templates. Feel free to reach out if there are any questions about that. Will have to check the deployed ui separately.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
ac-jorellanaf: I don't understand. Does this not pre-select the IVA?

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
ac-jorellanaf: It does not do anything in this case. The text size is being overridden by the Angular Material mixins.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
ac-jorellanaf: I thought there was a prior agreement on this? I changed it to make sure.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
ac-jorellanaf: This has been overhauled (and was my original desire but I ran out of time on Thursday).

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.ts
SilverLinings89: Sorry, I may be misunderstanding something here. If you look at this code in isolation, it is unclear why we set accessRequestIvaId and selectedIvaId (whats the difference?) and why we don't just this.#preSelectBestIva() from a purely linguistic standpoint. We should never be setting the not-best IVA if you understand what I mean. This is purely stylistic and we can leave it as-is. Just a remark on what I've noticed while reading this code. The naming just seems slightly sub-optimal but nothing that cannot be figured out by checking what it does.

SilverLinings89:
Hi Jordy, sorry, but did you make any changes? The function calls in templates are still there and Github says that the last commit was made 6 days ago.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
```diff
@@ -53,11 +55,11 @@ <h2 mat-dialog-title>Access Request Detail</h2>
             <tr>
               <td>
                 <mat-radio-button [value]="iva.id">
-                  {{ typeAndValue(iva) }}
+                  {{ (iva.type.toString() | IvaTypePipe: iva.value).typeAndValue }}
                 </mat-radio-button>
               </td>
-              <td class="{{ statusTextClass(iva) }} w-40">
-                {{ statusName(iva) }}
```
SilverLinings89: These are still function calls (toString() ) so this doesn't fix the problem yet.

src/app/verification-addresses/features/code-creation-dialog/code-creation-dialog.component.html
```diff
@@ -13,7 +13,7 @@ <h1 mat-dialog-title>Verification code created</h1>
       </tr>
       <tr>
         <th class="pe-1 text-right">Via:</th>
-        <td>{{ typeName(iva) }}</td>
+        <td>{{ (iva.type.toString() | IvaTypePipe).display }}</td>
```
SilverLinings89, line 16: Same problem here.

src/app/verification-addresses/features/user-iva-list/user-iva-list.component.html
```diff
@@ -22,7 +22,7 @@
     <table mat-table [dataSource]="ivaSource" class="mat-elevation-z8">
       <ng-container matColumnDef="value">
         <td mat-cell class="max-w-72 text-base" *matCellDef="let iva">
-          {{ typeAndValue(iva) }}
+          {{ (iva.type.toString() | IvaTypePipe).typeAndValue }}
```
SilverLinings89, line 25: see above

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
```diff
@@ -30,8 +32,12 @@
...
     <ng-container matColumnDef="state">
       <th mat-header-cell *matHeaderCellDef mat-sort-header>Status</th>
-      <td mat-cell class="{{ statusTextClass(iva) }}" *matCellDef="let iva">
-        {{ statusName(iva) }}
+      <td
+        mat-cell
+        class="{{ (iva.state.toString() | IvaStatePipe).class }}"
+        *matCellDef="let iva"
+      >
+        {{ (iva.state.toString() | IvaStatePipe).display }}
```
SilverLinings89, line 40: see above

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
```diff
@@ -12,13 +12,15 @@
     <ng-container matColumnDef="type">
       <th mat-header-cell *matHeaderCellDef mat-sort-header>Type</th>
       <td mat-cell *matCellDef="let iva">
-        <mat-icon class="mr-3">{{ typeIconName(iva) }}</mat-icon
-        ><span class="align-super">{{ typeName(iva) }}</span>
+        <mat-icon class="mr-3">{{ (iva.type.toString() | IvaTypePipe).icon }}</mat-icon
+        ><span class="align-super">{{
+          (iva.type.toString() | IvaTypePipe).display
```
SilverLinings89, line 17: see above

SilverLinings89 (changes requested):

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
ac-jorellanaf: We can't import models from a util. It is explicitly forbidden in ESLint, so we can't use the actual enum values for these things. We would have to recreate the enum in the pipe, but that is duplicating code.

Cito: If we find that some eslint rules for the architecture matrix turn are problematic, we can always discuss to relax or change something.

Cito: Let's discuss this case on the ui-ux channel.

Cito (dismissed):

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
SilverLinings89: Still here

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
```diff
@@ -1,54 +1,52 @@
...
       </td>
     </ng-container>
     <ng-container matColumnDef="state">
-      <th mat-header-cell *matHeaderCellDef class="w-40" mat-sort-header>Status</th>
-      <td mat-cell class="state{{ iva.state }}" *matCellDef="let iva">
-        {{ statusName(iva) }}
+      <th mat-header-cell *matHeaderCellDef mat-sort-header>Status</th>
+      <td
+        mat-cell
+        class="{{ (iva.state.toString() | IvaStatePipe).class }}"
```
SilverLinings89, line 37: function in template

src/app/verification-addresses/features/iva-manager-list/iva-manager-list.component.html
SilverLinings89: also here.

src/app/verification-addresses/features/user-iva-list/user-iva-list.component.html
SilverLinings89: still here.

SilverLinings89 (changes requested):
There are still some toStirngs(). Other than that fine for me.

Cito (approved):

Cito:
Looks like everything is addressed now, will merge.

PR #92, opened 2025-03-19
Improve runtime docker container

Cito:
- Remove buffer polyfill since it is very old, not an ESM, and not needed in the browser
- Remove warning because qrcode module is not an ESM
- Increase allowed bundle size to avoid warning (we may need to look into shrinking the bundle size though)
- Fix replacement of the version in ribbonText in prod mode and with production docker container
- Update some dependencies


package.json
```diff
@@ -28,36 +28,35 @@
...
-    "@angular/common": "^19.2.1",
-    "@angular/compiler": "^19.2.1",
-    "@angular/core": "^19.2.1",
-    "@angular/forms": "^19.2.1",
-    "@angular/material": "^19.2.2",
-    "@angular/material-date-fns-adapter": "^19.2.2",
-    "@angular/platform-browser": "^19.2.1",
-    "@angular/platform-browser-dynamic": "^19.2.1",
-    "@angular/router": "^19.2.1",
+    "@angular/animations": "^19.2.3",
```
SilverLinings89, line 31: Due to the ^ these all shouldn't have an effect

package.json
```diff
@@ -69,10 +68,10 @@
     "autoprefixer": "^10.4.21",
     "eslint": "^9.22.0",
     "eslint-config-prettier": "^10.1.1",
-    "eslint-import-resolver-typescript": "^3.8.5",
+    "eslint-import-resolver-typescript": "^4.2.2",
     "eslint-plugin-boundaries": "^5.0.1",
     "eslint-plugin-header": "^3.1.1",
-    "eslint-plugin-jsdoc": "^50.6.6",
```
SilverLinings89, line 75: This one has an effect

SilverLinings89 (approved):
All looks good to me. Minor remarks, nothing to fix. Was wondering about the complexity of the buffer replacement - want to look at it again tomorrow morning but at first glance I didn't understand it all and was wondering if we should test or document.

package.json
Cito: Yes, but I like to always have the latest versions in there just for info. It's also done automatically with "npm update".

Cito:
> was wondering if we should test or document.

Agreed, some of the methods in the component should be probably moved to utils and tested, added a note to the refactoring doc. So far I just ported the legacy code more or less 1:1.

Tested this manually: Entering a valid pub key works, entering a private key gives a specific validation warning.

PR #94, opened 2025-03-21
Added responsiveness to home page, header, and footer (GSI-1487)

ac-jorellanaf:
Added responsiveness to home page, header, and footer.

Fixed some minor style issues as well


Cito (dismissed):
Looks nice!

Only thing I noticed (maybe just on my monitor?) is that the social media links are very faint (close to white), they could be a bit more greyish (but not too much). You can also do that in a separate PR, since it probably has been like that before.

Cito (approved):
I think the e2e test still has the same problem.

Cito (approved):

Cito (dismissed):

Cito (approved):

PR #96, opened 2025-03-26
Fix bugs in rc.1

Cito:
This PR fixes two bugs that where discovered in 2.0.0-rc.1:
- When opening one dataset summary, the other ones where updated as well.
- A JavaScript error was thrown when sorting empty columns in the samples tables.

Also removed some duplication between the dataset summary and search result component. Both tried to load the data from the backend, and they always used the same instance of the service. Now only the search result component loads the data, every instance of the component uses its own service instance, and the result is passed on to the dataset summary component.


SilverLinings89 (dismissed):
LGTM. Did we change our linting rules? Because all the providedIn values are changed to one line ...

SilverLinings89 (approved):

PR #97, opened 2025-03-26
Fix bugs in rc.2

Cito:
This PR fixes some more bugs that have been found when testing rc.2 against the staging backend:

- The IVA types and values were not shown properly. This was caused by several problems with the implementation of the IvaTypePipe and IvaStatusPipe.
- The access requests were not properly sent to the backend; some fields were missing or in the wrong format.
- Also, some controls used both template driven and reactive form controls, causing a warning on the console.
- When there was an error on the backend, like due to the above, the success message was shown anyway.
- The IVA id was not submitted when allowing/denying requests. The cause was a signal that was not called.
- The backend does not allow reverting a status of "denied" to "allowed", so the frontend shouldn't offer that option either.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "data-portal",
-  "version": "2.0.0-rc.2",
+  "version": "2.0.0-rc.3",
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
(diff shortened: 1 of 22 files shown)


SilverLinings89 (approved):
LGTM

PR #98, opened 2025-04-01
Improve config for running e2e tests

Cito:
This PR adds a local docker-compose override config file with different volume mounts and DISPLAY settings depending on whether this is running under WSL or macOS.

Please test wether this runs on you machine:
- npm run e2e (headless e2e tests)
- npm run e2e:headed (headed e2e tests)
- npm run e2e:ui (e2e tests with graphical interface)
- npm run e2e:debug (e2e tests in debug mode)

The last three need an X11 server running on the host, which is included in WSL2 already.


ac-jorellanaf (approved):
Works perfectly for me in WSL in all cases

SilverLinings89 (approved):

PR #99, opened 2025-04-02
Responsive metadata browser (GSI-1431)

ac-jorellanaf:
Added responsiveness to metadata browser page and associated components.
Some minor corrections to comments for some pipe tests, etc.

### DIFF
--- a/src/app/metadata/features/metadata-browser/metadata-browser.component.html
+++ b/src/app/metadata/features/metadata-browser/metadata-browser.component.html
@@ -1,5 +1,5 @@
 <div class="grid grid-cols-12 grid-rows-none gap-x-5">
-  <div class="col-span-9 col-start-4">
+  <div class="col-span-12 xl:col-span-9 xl:col-start-4">
     <div class="flex w-full justify-between pb-4">
       <mat-chip-set aria-label="search and filters" class="filter-chips">
         @if (lastSearchQuery() !== '') {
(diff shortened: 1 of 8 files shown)


src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -116,9 +133,18 @@
           </button>
         </div>
       </form>
+      <div class="mt-6">
+        <button
+          mat-stroked-button
+          class="toggle-filters w-full"
+          (click)="displayFilters = !displayFilters"
+        >
+          <mat-icon>menu</mat-icon>Close filters
```
Cito, line 142: Since the other two buttons above do not have icons, this one directly below should also not have one. Alternatively, add an icon to all of them. But the "menu" icon looks a bit out of place here, particularly because the same button at the top of the section has different icon(s).

src/app/metadata/features/metadata-browser/metadata-browser.component.html
```diff
@@ -1,5 +1,5 @@
 <div class="grid grid-cols-12 grid-rows-none gap-x-5">
-  <div class="col-span-9 col-start-4">
+  <div class="col-span-12 xl:col-span-9 xl:col-start-4">
```
Cito, line 2: Maybe use the sidebar also with lg not just xl (but I'm not sure and let you decide).

src/app/metadata/features/metadata-browser/metadata-browser.component.html
Cito: Another problem is that the button also appears when this is displayed as sidebar, in which case it does not do anything.

Cito (changes requested):
Looks good to me, just some problems with the second "close filters" button, see below.

Cito (approved):
Looks good now. Except maybe that the "close filters" button now has different text on the top and on the bottom though it does exactly the same ("search and filters" vs just "filters").

I have added the idea to use a sidenav to our refactoring notes. But I think we should tackle that later after finishing the remaining responsiveness tasks.

ac-jorellanaf:
> Looks good now. Except maybe that the "close filters" button now has different text on the top and on the bottom though it does exactly the same ("search and filters" vs just "filters").

Since one has icons and the other does not, I thought it would be good to differentiate them.

PR #100, opened 2025-04-02
Added responsiveness to the dataset details page

ac-jorellanaf:


src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -274,8 +274,8 @@ <h4 class="inline-flex items-center">
                 [pageSize]="defaultTablePageSize"
                 [pageSizeOptions]="tablePageSizeOptions"
                 showFirstLastButtons
-                itemsPerPageLabel="Experiments per page"
                 aria-label="Select page of experiments"
+                class="sticky left-0 sm:static sm:left-[unset]"
```
Cito, line 278: Does anything speak against *always* using sticky left-0? Because even with larger than sm we still can have horizontal scrolling.

Cito (changes requested):
Looks good, just two minor things.

Cito (approved):

SilverLinings89 (approved):

PR #102, opened 2025-04-02
Various developer experience fixes

Cito:
- fix minor issues in the initialize.sh script
- improve the mock login process
- do not configure Umami by default (do not use it while developing unless explicitly configured in local.env)
- update the version to 2.0.0 (final release)


ac-jorellanaf (approved):
Looks good to me

PR #109, opened 2025-04-14
Fixes to accession mock strings and their displaying

ac-jorellanaf:
Fixed length of accession strings to match production
Fixed the way they display
Improved displaying of data not present in DINS
Fixed bottom spacing of buttons in dataset summary

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "data-portal",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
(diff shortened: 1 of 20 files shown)


Cito (changes requested):
The accessions  (GHGA and EGA) now have one digit more than in production, is this by intent?

Also, the DINS results do not match any more because of the changed GHGAF accessions, therefore the total size does not match and the e2e test fails.

I suggest to change it like it was before: 5 process data files in HD with MB size, 5 research data files in TUE with GB size, no info for the 2 supporting files.

Cito (approved):

PR #110, opened 2025-04-14
Add dataset ID to download token

Cito:
The full download token also contains the dataset ID as a prefix, but we only showed the actual token itself.

This is now changed so that the full download token (as it needs to be pasted to the connector) is displayed and copied to the clipboard.

The PR also updates all dependencies and bumps the patch version.


dontseyit (approved):
Looks good to me

PR #111, opened 2025-04-14
Add proper ID to download token

Cito:
The download token should be prefixed with the internal work package ID, not with the dataset ID.

This PR also makes sure that the hyphens in the ID (a UUID) do not cause a word wrap.


dontseyit (approved):
Thanks a lot

PR #112, opened 2025-04-22
Hotfix for publications and pagination

ac-jorellanaf:
Fixes mismatch of optional parameters in Publication model between th…e data portal and ground truth.
Fixes pagination behaviour when filtering.
Bumps version to 2.0.4.

### DIFF
--- a/src/app/metadata/features/dataset-details/dataset-details.component.html
+++ b/src/app/metadata/features/dataset-details/dataset-details.component.html
@@ -140,15 +140,17 @@
                   <p class="text-base"><strong>Year:</strong> {{ pub.year }}</p>
                 }
                 <p class="text-base"><strong>DOI:</strong> {{ pub.doi }}</p>
-                @for (
-                  par of pub.abstract?.split('\\n');
-                  track $index;
-                  let first = $first
-                ) {
-                  <p class="text-base">
-                    <strong>{{ first ? 'Abstract: ' : '' }}</strong
-                    >{{ par }}
-                  </p>
+                @if (pub.abstract) {
+                  @for (
+                    par of pub.abstract.split('\\n');
+                    track $index;
+                    let first = $first
+                  ) {
+                    <p class="text-base">
+                      <strong>{{ first ? 'Abstract: ' : '' }}</strong
+                      >{{ par }}
+                    </p>
+                  }
                 }
                 @if (!last) {
                   <hr class="mx-14 my-6" />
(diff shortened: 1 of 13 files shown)


src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -128,13 +128,20 @@ <h4 class="inline-flex items-center">
...
+                @if (pub.journal) {
+                  <p class="text-base"><strong>Journal:</strong> {{ pub.journal }}</p>
+                }
+                @if (pub.year) {
+                  <p class="text-base"><strong>Year:</strong> {{ pub.year }}</p>
+                }
+                <p class="text-base"><strong>DOI:</strong> {{ pub.doi }}</p>
                 @for (
-                  par of pub.abstract.split('\\n');
+                  par of pub.abstract?.split('\\n');
```
Cito, line 144: I would wrap this for loop into an `if (pub.abstract)` because it's an implementation detail how the Angular compiler deals with an undefined value in this place, I would not want to rely on that (in normal JavaScript, you cannot iterate over undefined).

Ideally, paragraphs should be also trimmed and empty paragraphs be filtered out, but not in this hotfix because this probably should be done in a pipe.

Cito (changes requested):

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -156,6 +158,7 @@ export class MetadataBrowserComponent implements OnInit {
   clearSearchQuery(): void {
     this.searchTerm = '';
     this.searchFormControl.setValue('');
+    this.#skip = 0;
     this.#performSearch();
```
Cito, line 162: if (this.searchTerm) {
      this.searchTerm = '';
      this.searchFormControl.setValue('');
      this.#skip = DEFAULT_SKIP_VALUE;
      this.#performSearch();
    }

Cito (changes requested):

src/app/metadata/features/dataset-details/dataset-details.component.html
ac-jorellanaf: Now that I think about it, would this not be one of these 'no function calls in template'? @SilverLinings89

Cito: Yes, I think we should either precompute the array and store it in a property or use a pipe. Doing so, we could then also do the trimming and filtering.

ac-jorellanaf: We should do this in a different PR.

Cito (approved):

Cito (approved):

PR #117, opened 2025-04-25
Added better error handling for both the browse page and dataset details page (GSI-1554)

ac-jorellanaf:


src/app/metadata/pipes/parse-error.pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+ * Transforms an HttpRequest error signal into a readable message.
+ */
+@Pipe({
+  name: 'parseError',
+})
+export class ParseErrorPipe implements PipeTransform {
+  /**
+   * The transform method executes the business logic of the Pipe
+   * @param error A signal containing the error object.
+   * @returns A string containing the error message or an empty string if the error is in the JSON parsing (which gives an HTTP200 status code)
```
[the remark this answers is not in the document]
ac-jorellanaf: It does. If there is an error in the JSON, the HttpResponse is 200, but there is still an error during the parsing of the JSON.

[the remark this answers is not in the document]
Cito: Yes, that can happen, but then this should be treated as an error and the user should see a general error message, same as a 500 error. My suggested code snippet above would handle that.

src/app/metadata/features/dataset-details/dataset-details.component.ts
```diff
@@ -87,10 +85,21 @@ export class DatasetDetailsComponent implements OnInit, AfterViewInit {
   #dins = inject(DatasetInformationService);
 
   #datasetDetails = this.#metadata.datasetDetails;
-  error = this.#datasetDetails.error;
+  #error = this.#datasetDetails.error;
   datasetDetails = this.#datasetDetails.value;
   datasetInformation = this.#dins.datasetInformation.value;
 
+  errorMessage = computed(() => {
+    switch (this.#error()) {
```
Cito, line 93: Shouldn't this be `#error().status?`.

Also, I would use a more specific name like `#datasetDetailsError` (there could also be an error for the dins).

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -69,10 +67,7 @@ export class MetadataBrowserComponent implements OnInit {
   lastSearchQuery = this.#metadataSearch.query;
   lastSearchFilterFacets = this.#metadataSearch.facets;
   #searchResults = this.#metadataSearch.searchResults;
-  error = computed(() => {
-    console.log(this.#searchResults.error());
-    return this.#searchResults.error();
-  });
+  #error = this.#searchResults.error;
```
Cito, line 70: see above

src/app/metadata/features/metadata-browser/metadata-browser.component.ts
```diff
@@ -83,6 +78,15 @@ export class MetadataBrowserComponent implements OnInit {
   numResults = computed(() => this.searchResults().count);
   loading = computed(() => this.#searchResults.isLoading());
 
+  errorMessage = computed(() => {
+    switch (this.#error()) {
+      case undefined:
+        return undefined;
+      default:
+        return 'There was an error loading the datasets. Please try again later.';
```
Cito, line 86:
```suggestion
        return 'There was an error browsing the datasets. Please try again later.';
```

src/app/metadata/features/search-result-list/search-result-list.component.html
```diff
@@ -1,7 +1,7 @@
-<div class="bg-primary mb-2 grid w-full grid-cols-[14rem_1fr] px-6 py-2 text-white">
+<div class="bg-primary mb-4 grid w-full grid-cols-[14rem_1fr] px-6 py-2 text-white">
   <span>Dataset ID</span><span>Title</span>
 </div>
-<div [class.invisible]="!loading()" class="mb-3">
+<div [class.hidden]="!loading()" class="mb-3">
```
Cito, line 4: I think the "invisible" was intentional, so that after loading the layout would not change (the results would appear exactly where the skeletons were).

Cito (changes requested):

src/app/metadata/features/search-result-list/search-result-list.component.html
ac-jorellanaf: This should only be the loading bar? The problem is with invisible, the div causes spacing issues.

src/app/metadata/features/dataset-details/dataset-details.component.ts
ac-jorellanaf: You're right! I will fix it

src/app/metadata/features/search-result-list/search-result-list.component.html
Cito: Can you fix this in a separate PR?

Cito (approved):

Cito (approved):

PR #118, opened 2025-04-29
Implemented calls to the Well-Known Values Service for the human-readable storage aliases (GSI-1565)

ac-jorellanaf:
Removed the pipe hardcoding the aliases.
In search for a better object name
Pending possible change for API path


src/app/metadata/services/well-known.service.ts
Cito, line 1: I would name this file well-known-value-service.ts, same for spec

src/app/metadata/services/well-known.service.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Well-Known Value query service
+ *
+ * This service provides the functionality to fetch well-known values.
+ */
+@Injectable({ providedIn: 'root' })
+export class WellKnownValueService {
+  #config = inject(ConfigService);
+  #wkvsUrl = this.#config.wkvsUrl;
+
+  #storageAliasDecodesUrl = `${this.#wkvsUrl}/storage_alias_decodes`;
```
Cito, line 25: I think the path is wkvsUrl/values/storage_labels (note the /values/ part).

src/app/metadata/models/storage-aliases.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Interface for the storage aliases.
+ * @copyright The GHGA Authors
+ * @license Apache-2.0
+ */
+
+export interface storageAliasDecodes {
```
Cito, line 7: actually this type applies to all values returned by the wkvs, so you could call it something like `WellKnownValues` (names of interfaces should start with a capital letter)

src/app/metadata/models/storage-aliases.ts
Cito, line 1: this file should be better called well-known-values.ts

src/mocks/data.ts
```diff
@@ -646,3 +647,18 @@ export const workPackageResponse: WorkPackageResponse = {
   token:
     'gumRa5XE1Rm3tOJr3ocfA1F63rRUI2g6eCI0KY2Mv3epb28cZeylvaYsxRmgQRDboE2yOhtE4qxPhZgYz/Y7zR+hssBzq7Hg',
 };
+
+export const storageAliasDecodes: baseStorageAliasDecodes = {
```
Cito, line 651:
```suggestion
export const storageAliasDecodes: BaseStorageAliasDecodes = {
```

type names should be uppercase

src/mocks/data.ts
```diff
@@ -9,6 +9,7 @@ import { DatasetInformation } from '@app/metadata/models/dataset-information';
 import { DatasetSummary } from '@app/metadata/models/dataset-summary';
 import { BaseGlobalSummary } from '@app/metadata/models/global-summary';
 import { SearchResults } from '@app/metadata/models/search-results';
+import { BaseStorageAliasDecodes as baseStorageAliasDecodes } from '@app/metadata/models/storage-aliases';
```
Cito, line 12:
```suggestion
import { BaseStorageAliasDecodes } from '@app/metadata/models/storage-aliases';
```

src/app/metadata/models/storage-aliases.ts
```diff
@@ -0,0 +1,15 @@
...
+  [key: string]: string;
+}
+
+export interface BaseStorageAliasDecodes {
+  storage_alias_decodes: storageAliasDecodes;
```
Cito, line 12: this has been renamed to `storage_labels` in wkvs

src/app/metadata/services/well-known.service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * The human-readable storage aliases (empty while loading) as a resource
+   */
+  storageAliasDecodes = httpResource<storageAliasDecodes>(
+    this.#storageAliasDecodesUrl,
+    {
+      parse: (raw) => (raw as { alias_decodes: storageAliasDecodes }).alias_decodes,
```
Cito:
```suggestion
      parse: (raw) => (raw as BaseStorageAliasDecodes).alias_decodes,
```

src/app/metadata/services/well-known-service.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        provideHttpClient(),
+        provideHttpClientTesting(),
+        { provide: ConfigService, useClass: MockConfigService },
+        { provide: WellKnownValueService },
+      ],
+    });
+    service = TestBed.inject(WellKnownValueService);
+    httpMock = TestBed.inject(HttpTestingController);
+  });
+
```
Cito, line 44: You should add a  httpMock.verify() in afterEach as in the other tests.

Cito (changes requested):
There was another name change in the wkvs, see also suggestions below.

src/app/metadata/services/well-known.service.ts
ac-jorellanaf: No, because WKVS Url already contains /values/

Cito: Makes sense, but for consistency I think it's better to change this. All our other URLs in the settings refer to the base addresses of the respective service, even if the service has only one resource/endpoint. And in archive test bed the `wkvs_url` config setting is also the base address, `/.well-known`.

In the service itself, you can add an internal property `#values_url` that would be `#wkvs_url` + `/values`. Similarly to how we do it with `#workPackagesUrl` and `#usersUrl` in the work-package.service.

src/app/metadata/features/dataset-details/dataset-details.component.html
```diff
@@ -442,7 +442,12 @@ <h4 class="lg:max-w-auto inline-flex items-center md:max-w-[26rem]">
                     File location
                   </th>
                   <td mat-cell *matCellDef="let file">
-                    {{ (file.file_information?.storage_alias | storageAlias) || 'N/A' }}
+                    @let storageAlias = file.file_information?.storage_alias;
+                    @if (storageAlias) {
+                      storageLabels()[storageAlias] || storageAlias
+                    } @else {
+                      'N/A'
```
Cito, line 449:
```suggestion
                    @if (storageAlias) {
                      {{ storageLabels()[storageAlias] || storageAlias }}
                    } @else {
                      N/A
```

Cito (changes requested):

src/app/metadata/features/dataset-details/dataset-details.component.html
Cito: Or, use a ternary expression again inside double braces.

Cito (approved):

SilverLinings89 (approved):

PR #119, opened 2025-05-12
Update Angular, pnpm and other requirements

Cito:


SilverLinings89 (approved):
Looks good and runs without any difference (checked ui, unit tests and installation) for me.

ac-jorellanaf (approved):
Everything looks good for me. E2E tests, unit tests, the behaviour of the portal (both with an without backend) works.

PR #122, opened 2025-05-14
Add a component to show notes. GSI-1582

SilverLinings89:
Beginning work on the updated access requests user experience. Notes are now being shown for both users and data stewards.


src/app/access-requests/features/access-request-note/access-request-note.component.html
ac-jorellanaf, line 1: I feel like using a separate component for this is a little unnecessary. It's easy to tell which type of note should be displayed based on context. IVA manager = both, account page = to reviewer. You're not really saving any code with this.

On the other hand, proper unit testing for the various options is a welcome addition.

ac-jorellanaf (changes requested):

src/app/access-requests/features/access-request-note/access-request-note.component.html
SilverLinings89: But I fully agree - in the current form it seems like an early refactoring.

ac-jorellanaf: Well, there is only one place where the to requester note should be visible and edited (AR manager). And maybe only one place (2 in the future if we let users manage their ARs too) where the to reviewer should be visible and edited (request access dialog). So it doesn't seem like the reusability is much of an issue here. Component bloat is for certain though, though we already have components that are overflowing (DS details page, browse component).
Testing is also pretty good though, and if you are planning to make the notes visible only and only editable after pressing a button, then I'm 100% on separating the component.

ac-jorellanaf (dismissed):

PR #123, opened 2025-05-14
Display expiration of the access request when selecting dataset

ac-jorellanaf:


src/app/work-packages/features/work-package/work-package.component.ts
```diff
@@ -21,7 +21,7 @@ import { RouterLink } from '@angular/router';
 import { NotificationService } from '@app/shared/services/notification.service';
 import { FRIENDLY_DATE_FORMAT } from '@app/shared/utils/date-formats';
 import { getBackendErrorMessage, MaybeBackendError } from '@app/shared/utils/errors';
-import { Dataset } from '@app/work-packages/models/dataset';
+import { DatasetWithExpiration } from '@app/work-packages/models/dataset';
```
SilverLinings89, line 24: I don't think this naming is good. I see where you are going with it but I would normally make one distinction. If you have an object { dataset: Dataset, expiration: Date} then the name would be good. However, that's not the case here. The expiration is part of the Dataset type so there is no Dataset without an expiration.

SilverLinings89 (changes requested):

src/app/work-packages/features/work-package/work-package.component.ts
Cito: I think the name is good, as it's based on the OpenAPI spec. The WPS returns a DatasetWithExpiraton object that contains fields describing the dataset itself plus a field describing when _access_ to the dataset (for the particular user who requested the list of datasets from the WPS) expires. This field does not belong to the dataset itself.

Cito: Of course you can argue that the backend API should have nested the dataset instead of creating a subtype with all of the fields. But I don't think we should change that.

Cito (dismissed):
Looks ok to me.

Please try to keep the length of the PR title under 72 chars (ideally < 50) though, in order to keep the list of PRs concise and because it's used as commit message which have similar rules to make the commit log better readable.

If you need to give more details, put them in the description of the PR, which can become the description of the commit as well.

src/app/work-packages/features/work-package/work-package.component.ts
ac-jorellanaf: Done now

SilverLinings89 (approved):

Cito (approved):

PR #125, opened 2025-05-15
Added ticket ID and notes to request filtering

ac-jorellanaf:


Cito (dismissed):
Thank you, this looks nice.

I know it's not part of the original task, but would it make sense to also make the description (request text) filterable? Then we would have an even nicer 3x3 grid of filters with the right column for the different kinds of notes.

src/app/access-requests/services/access-request.service.ts
```diff
@@ -205,6 +207,10 @@ export class AccessRequestService {
       if (filter.status) {
         requests = requests.filter((ar) => ar.status === filter.status);
       }
+      const requestText = filter.requestText?.trim().toLowerCase();
+      if (requestText) {
+        requests = requests.filter((ar) => ar.request_text === requestText);
```
Cito, line 212: This should work similar to the other notes (includes instead of equal).

Cito (changes requested):

src/app/access-requests/services/access-request.service.ts
ac-jorellanaf: You're right. And so should the ticket ID!

Cito (approved):

PR #128, opened 2025-05-19
Global summary info from all files (GSI-1632)

ac-jorellanaf:


src/app/metadata/features/global-summary/global-summary.component.ts
```diff
@@ -47,5 +48,25 @@ export class GlobalSummaryComponent {
   datasets = computed(() => this.stats().Dataset);
   methods = computed(() => this.stats().ExperimentMethod);
   individuals = computed(() => this.stats().Individual);
-  files = computed(() => this.stats().ProcessDataFile);
+  files: Signal<FileStatsModel> = computed(() => {
+    let fileCount = 0;
+    const formats: { value: string; count: number }[] = [];
```
Cito, line 53: It would be better to make `formats` a `Map` (from value to count). Then you can avoid the slow find operations, and use gets instead. At the end, you can convert the Map back to the array of value/count objects.

Cito (changes requested):
Looks good, just some suggestions to make it a bit cleaner and faster.

src/app/metadata/features/global-summary/global-summary.component.ts
```diff
@@ -47,5 +48,34 @@ export class GlobalSummaryComponent {
...
+          }
+        });
+      });
+    const formatsDict: ValueCount[] = Array.from(formats.entries()).map(
+      ([value, count]) => ({ value, count }),
+    );
+    return {
+      count: fileCount,
+      stats: formatsDict.length > 0 ? { format: formatsDict } : undefined,
+    };
```
Cito, line 79:
```suggestion
    let totalCount = 0;
    const totalFormats: Map<string, number> = new Map();
    Object.entries(this.stats())
      .filter(([key]) => key.endsWith('File'))
      .forEach(([, value]) => {
        const fileStats = value as FileStats;
        totalCount += fileStats.count;
        const format = fileStats.stats?.format || [];
        format.forEach(({ value, count }) =>
          totalFormats.set(value, (totalFormats.get(value) || 0) + count),
        );
      });
    const format: ValueCount[] = Array.from(totalFormats.entries()).map(
      ([value, count]) => ({ value, count }),
    );
    return {
      count: totalCount,
      stats: format ? { format } : undefined,
    };
```

I simplified it a bit, made the names a bit clearer and added a check for the case that fileStats.stats is undefined.

Cito (approved):

PR #131, opened 2025-05-21
Merge 2.0.x maintenance branch back to main

Cito:


src/app/metadata/features/global-summary/global-summary.component.ts
```diff
@@ -47,5 +48,31 @@ export class GlobalSummaryComponent {
...
+        );
+      });
+    const format: ValueCount[] = Array.from(totalFormats.entries()).map(
+      ([value, count]) => ({ value, count }),
+    );
+    return {
+      count: totalCount,
+      stats: format ? { format } : undefined,
+    };
+  }
```
SilverLinings89, line 77: Maybe testing this would be nice. Not required, though.

SilverLinings89 (approved):
Two somewhat length functions that might be good to test but fine as is.

src/app/metadata/features/global-summary/global-summary.component.ts
Cito: Agree. But this is a private method. We could need to either test it by looking at the component output or make it a public utility function. Can do that in a separate PR.

PR #132, opened 2025-05-21
Add DAC alias and dataset title to the access request details view

SilverLinings89:
Also turn dataset id and requester email into links.

See [GSI-1644](https://youtrack-ghga.dkfz.de/issue/GSI-1644/data-portal-Show-DAC-and-dataset-title-in-access-request-details)


src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
```diff
@@ -21,7 +21,22 @@ <h2 mat-dialog-title>Access Request Detail</h2>
       }
       <tr>
         <th class="pe-1 text-start">Dataset ID:</th>
-        <td>{{ data.dataset_id }}</td>
+        <td>
+          <a
+            class="underline"
```
Cito, line 26: The underline for links is actually included in the global style sheet.

The problem is that the selector there only looks for links in the main element, but the dialog is a separate container.

I think we should fix this in the global style sheet and remove the explicit underlines.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
```diff
@@ -31,14 +46,27 @@ <h2 mat-dialog-title>Access Request Detail</h2>
...
         </td>
       </tr>
-
+      <tr>
+        <th class="pe-1 text-start align-top">DAC:</th>
+        <td>{{ data.dac_alias }}</td>
+      </tr>
+      <tr></tr>
+      <tr>
+        <th class="pe-1 text-start align-top">Dataset Title:</th>
```
Cito, line 55: We should consistently use either title case or sentence case for the labels in the THs. Currently it's a mix.

Cito (dismissed):
Looks good, just two small remarks.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
SilverLinings89: fixed.

src/app/access-requests/features/access-request-manager-dialog/access-request-manager-dialog.component.html
SilverLinings89: fixed.

Cito (approved):

PR #136, opened 2025-05-27
Adding start and end date edit functionality to Access Request Dialog

SilverLinings89:


ac-jorellanaf (changes requested):
The dates should be edited at the same time, rather than individually.

Cito (approved):
I'd also prefer having both fields under one edit button.

You could use a virtual field name "period" that the field edit component would be aware of and then display these two fields instead of just one.

In the unedited state, the fields could appear on the same line with a label like "validity period" and then the start and end dates. The start can always be update to now() if it's in the past since this is done by the backend as well when values are changed.

In the edited states, they could be shown as two dates.

However, this would also require a change in the signature of the "saved" and "edited" methods (provided as output events). Currently they can handle only the case that a single field is edited.

Also, the dates should be validated so that you cannot enter a start date in the past or after the end date. This should be handled similarly to the access request dialog where we already have such validation.

I approved so that you can merge already and move the rest to a separate (new) task/PR.

PR #149, opened 2025-06-03
Improve IVA verification process (GSI-1667)

Cito:
This PR is an attempt to solve the issue described [here](https://github.com/ghga-de/forum/discussions/61).

Instead of showing a static text "Verification requested", we now show a button that says "Awaiting verification". When the button is clicked, it checks whether the code has been created, by force reloading the IVAs. If yes, it asks the user to enter it. If no, it tells the user to wait and retry later.

Update: We changed this. The table now shows two columns, one for the state and one for the corresponding action button. The button will be labeled "Enter verification code" in both cases.


Cito:
I have added a separate column for the buttons now. The layout and responsiveness still needs to be fixed.

PR #150, opened 2025-06-04
Simultaneous access duration dates editing (GSI-1658)

ac-jorellanaf:
Initial implementation, probably still not perfect.
Extended the edit field component to allow for multiple inputs simultaneously.
Minor fixes in the site footer and the AR dialog.

### DIFF
--- a/src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.html
+++ b/src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.html
@@ -1,86 +1,90 @@
-@if (!locked() || field()) {
-  <tr>
-    <th
-      class="w-36 min-w-36 pe-1 text-start"
-      [class.align-top]="!isOpen()"
-      [class.align-center]="isOpen()"
-    >
-      {{ label() }}:
-    </th>
-    <td class="flex min-w-60 items-center" [class.pb-[2px]]="!isOpen()">
-      @if (isOpen()) {
-        @if (isDate()) {
-          <mat-form-field class="grow">
-            <mat-datepicker-toggle
-              matIconSuffix
-              class="float-end"
-              [for]="datepicker"
-            ></mat-datepicker-toggle>
-            <input
-              matInput
-              [matDatepicker]="datepicker"
-              [(ngModel)]="field"
-              (dateChange)="onDateSelected($event)"
-              class="pr-8"
-            />
-
-            <mat-datepicker #datepicker></mat-datepicker>
-          </mat-form-field>
-        } @else {
-          <mat-form-field class="grow">
-            @if (rows() > 1) {
-              <textarea
+@if (!locked() || names()) {
+  @for (name of names(); track name; let idx = $index) {
+    <tr>
+      <th
+        class="w-36 min-w-36 pe-1 text-start"
+        [class.align-top]="!isOpen()"
+        [class.align-center]="isOpen()"
+      >
+        {{ labels()[idx] ? labels()[idx] + ':' : '' }}
+      </th>
+      <td class="flex min-w-60 items-center" [class.pb-[2px]]="!isOpen()">
+        @if (isOpen()) {
+          @if (isDate()) {
+            <mat-form-field class="grow">
+              <mat-datepicker-toggle
+                matIconSuffix
+                [for]="datepicker"
+              ></mat-datepicker-toggle>
+              <input
                 matInput
-                [(ngModel)]="field"
-                (ngModelChange)="changed()"
-                [rows]="rows()"
-              ></textarea>
-            } @else {
-              <input matInput [(ngModel)]="field" (ngModelChange)="changed()" />
-            }
-          </mat-form-field>
-        }
-        <mat-chip
-          class="smaller-chip interactive-chip cancel-edit-button mx-2"
-          (click)="cancel()"
-        >
-          <mat-icon>close</mat-icon>
-        </mat-chip>
-        <mat-chip
-          class="smaller-chip interactive-chip save-edit-button"
-          (click)="save()"
-        >
-          <mat-icon>check</mat-icon>
-        </mat-chip>
-      } @else {
-        <span class="grow">
-          @if (ticketUrl()) {
-            <a [href]="ticketUrl()" target="_blank" tabindex="-1"
-              >{{ field()
-              }}<mat-icon
-                role="img"
-                aria-hidden="true"
-                class="mat-icon material-symbols-outlined mat-icon-no-color no-translate h-4 w-4"
-                data-mat-icon-type="font"
-                >open_in_new</mat-icon
-              ></a
-            >
+                [matDatepicker]="datepicker"
+                [(ngModel)]="fields()[idx]"
+                (dateChange)="onDateSelected($event, idx)"
+                class="pr-8"
+              />
+              <mat-datepicker #datepicker></mat-datepicker>
+            </mat-form-field>
           } @else {
-            @if (isDate()) {
-              {{ field() | date: 'yyyy-MM-dd' }}
-            } @else {
-              {{ field() }}
-            }
+            <mat-form-field class="grow">
+              @if (rows()[idx] > 1) {
+                <textarea
+                  matInput
+                  [(ngModel)]="fields()[idx]"
+                  (ngModelChange)="changed(idx)"
+                  [rows]="rows()"
+                ></textarea>
+              } @else {
+                <input
+                  matInput
+                  [(ngModel)]="fields()[idx]"
+                  (ngModelChange)="changed(idx)"
+                />
+              }
+            </mat-form-field>
           }
-        </span>
-        @if (!locked()) {
           <mat-chip
-            (click)="edit()"
-            class="smaller-chip interactive-chip edit-button ms-2"
-            ><mat-icon class="scale-[0.75]">edit</mat-icon>
+            class="smaller-chip interactive-chip cancel-edit-button mx-2 mb-5"
+            (click)="cancel()"
+          >
+            <mat-icon>close</mat-icon>
           </mat-chip>
+          <mat-chip
+            class="smaller-chip interactive-chip save-edit-button mb-5"
+            (click)="save()"
+          >
+            <mat-icon>check</mat-icon>
+          </mat-chip>
+        } @else {
+          <span class="grow">
+            @if (ticketUrl()) {
+              <a [href]="ticketUrl()" target="_blank" tabindex="-1"
+                >{{ fields()[idx]
+                }}<mat-icon
+                  role="img"
+                  aria-hidden="true"
+                  class="mat-icon material-symbols-outlined mat-icon-no-color no-translate h-4 w-4"
+                  data-mat-icon-type="font"
+                  >open_in_new</mat-icon
+                ></a
+              >
+            } @else {
+              @if (isDate()) {
+                {{ fields()[idx] | date: 'yyyy-MM-dd' }}
+              } @else {
+                {{ fields()[idx] }}
+              }
+            }
+          </span>
+          @if (!locked()) {
+            <mat-chip
+              (click)="edit()"
+              class="smaller-chip interactive-chip edit-button ms-2"
+              ><mat-icon class="scale-[0.75]">edit</mat-icon>
+            </mat-chip>
+          }
         }
-      }
-    </td>
-  </tr>
+      </td>
+    </tr>
+  }
 }
--- a/src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
+++ b/src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
@@ -8,10 +8,10 @@ import { DatePipe } from '@angular/common';
 import {
   Component,
   computed,
-  effect,
   inject,
   input,
   model,
+  OnInit,
   output,
   signal,
 } from '@angular/core';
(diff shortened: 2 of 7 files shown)


Cito (changes requested):
The main logic looks good to me, but I have to major requests:

**1. The dates need validation:**

The validation should work similarly to the access request dialog.

The *start date* should be always pushed forward to the current date if it is in the past (when opening the editor), and it should not be possible to select any date that is in the past.

The *end date* must be validated to be larger than the start date. The access request service also has an upper limit for the end date. This is controlled via the setting `access_grant_max_extend` which should be added to the front end as well, so that we can validate it on the front end side, too.

You can create a separate task / PR for the validation if you like. But the idea was to run the validation (on the front end) and show the errors when the small "save" button is clicked.

**2. Duplicate buttons:**

The two buttons that do the same thing are confusing for users. There should be only one button (or save/cancel button pair) for both fields. The simplest solution would be to show the button(s) only for the last field in the field editor. Maybe you can draw a light border around the two fields so that the user understands they belong together.

src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
```diff
@@ -38,58 +47,249 @@ import { ConfigService } from '@app/shared/services/config.service';
...
+    this.fieldErrorMessages.set(Array(this.names().length).fill(''));
+  }
+
+  /**
+   * Update the range for the until date based on the from date
+   * @param date - The from date to update the range for
+   */
+  updateAccessEndRanges(date: Date): void {
+    const newUntilMinDate = new Date(
+      date.getTime() + this.#config.accessGrantMinDays * MILLISECONDS_PER_DAY,
```
Cito, line 145: Since our config values are given as number of days, I would use calculations based on days instead. Something like:
```js
const maxDate = new Date();
maxDate.setDate(date.getDate() + maxDays);
```
Could also avoid problems with leap seconds or DST.

src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
```diff
@@ -38,58 +47,249 @@ import { ConfigService } from '@app/shared/services/config.service';
...
+          Validators.required,
+          (control) => this.#untilDateValidator(control, idx),
+        ]);
+      return null;
+    });
+    this.fieldErrorMessages.set(Array(this.names().length).fill(''));
+  }
+
+  /**
+   * Update the range for the until date based on the from date
```
Cito, line 140: Since you're calculating the limits for the field based on the other field, we lose the (only) advantage of having a simultaneous editor for both fields, namely that the period may be temporarily invalid while editing, and the validation only happens when you click "save".

E.g. assume the max period is 1 years. Currently the period is start: now, end: in one year. Assume you now want to postpone that period for one year, i.e. start: in one year, end: in two years. If you change either start or end date like that, the period would be temporarily invalid until you set the other date.

Probably it's not a big problem in practice, and the user has ways to work around by moving the period in smaller steps. But I wonder if that's the case and we do validation for the individual fields anyway, then we might just as well revert back to the original implementation where you edit the fields individually.

src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.html
```diff
@@ -1,86 +1,107 @@
...
-
-            <mat-datepicker #datepicker></mat-datepicker>
-          </mat-form-field>
-        } @else {
-          <mat-form-field class="grow">
-            @if (rows() > 1) {
-              <textarea
+@if (!locked() || names()) {
+  @for (name of names(); track name; let idx = $index; let first = $first) {
+    <tr class="">
```
Cito, line 3:
```suggestion
    <tr>
```

src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
```diff
@@ -38,58 +47,249 @@ import { ConfigService } from '@app/shared/services/config.service';
...
+   * @param date - The until date to update the range for
+   */
+  updateAccessStartRanges(date: Date): void {
+    const currentDate = this.todayMidnight;
+
+    const newFromMinDate = new Date(
+      Math.max(
+        currentDate.getTime(),
+        date.getTime() - this.#config.accessGrantMaxDays * MILLISECONDS_PER_DAY,
+      ),
```
Cito, line 166: Here you're applying the (stricter) restrictions for the _users_ when doing access requests. The _data stewards_ have less strict restrictions:

- start can be today or later, `accessUpfrontMaxDays` is not enforced in this case
- duration must be `accessGrantMaxDays * accessGrantMaxExtend` days at max

See our implementation in the access request service. Note the additional factor `accessGrantMaxExtend` that allows the data stewards to grant longer time periods than users can request if needed. This still needs to be added as config setting to the frontend. Currently the setting is only used in the access request service.

Cito (changes requested):
This PR shows that in practice the simultaneous date editing is less straightforward than thought.

The edit buttons for the dates that appear only for the first field are a bit confusing, since there is no visual grouping of the edited fields and usually the data stewards wants to edit the end date, not the start date.

However, I understand that creating a visual grouping is a difficult because of the chosen design using tables with flex boxes inside cells.

See my suggestions for solutions: Vertical align the buttons at the top. Also consider reverting to editing individual fields, particularly if you're validating the fields individually anyway (as you're currently doing). Since then I see not much advantage in the simultaneous editing.

Also note that the exact validation rules for data stewards are a bit different from the validation rules for requesters. In case of doubt, always check the implementation in the backend. The frontend should always duplicate the validation there.

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+    const ret = this.#validateDateAgainstMinAndMax(
+      control.value,
+      this.minFromDate,
+      this.maxFromDate,
+    );
+    this.fromFieldErrorMessage.set(
+      this.getErrorMessageForValidationState('start', ret),
+    );
+    return ret;
+  };
```
SilverLinings89, line 238: I think you can merge these two (just a draft)
```suggestion
  #DateValidator = (
    control: AbstractControl<Date, Date>,
    from: boolean,
  ): ValidationErrors | null => {
const [min, max, field, name] = from 
  ? [this.minFromDate, this.maxFromDate, this.fromFieldErrorMessage, 'start'] 
  : [this.minUntilDate, this.maxUntilDate, this.untilFieldErrorMessage, 'end'];
    const ret = this.#validateDateAgainstMinAndMax(
      control.value,
      min,
      max,
    );
    field.set(
      this.getErrorMessageForValidationState(name, ret),
    );
    return ret;
  };
```

SilverLinings89: Also: Logic like this should be tested.

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  edit = () => {
+    this.isOpen.set(true);
+  };
+
+  onDateSelected = (
+    selectedDate: MatDatepickerInputEvent<Date> | Date,
+    name: string,
+  ) => {
+    let value: Date | null;
```
SilverLinings89, line 265:
```suggestion
    let value: Date | null = null;
```
Otherwise the variable is actually undefined after this line which is neither null nor Date.

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+
+  todayMidnight = new Date();
+  tomorrowMidnight = new Date();
+  minFromDate = new Date();
+  minUntilDate = new Date();
+  maxFromDate = new Date();
+  maxUntilDate = new Date();
+
+  fromFieldErrorMessage = signal<string>('');
+  untilFieldErrorMessage = signal<string>('');
```
SilverLinings89, line 80: Defining these as string | null and setting them to null if there is no error would be clearer. That way, you can then check if(this.fromFieldErrorMessage()) (or with !) instead of comparing against an empty string.

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+  onDateSelected = (
+    selectedDate: MatDatepickerInputEvent<Date> | Date,
+    name: string,
+  ) => {
+    let value: Date | null;
+    if (selectedDate instanceof MatDatepickerInputEvent) value = selectedDate.value;
+    else value = selectedDate;
+    if (value) {
+      const selectedLocalDate = new Date(value);
+      if (name === 'access_starts') {
```
SilverLinings89, line 270: as mentioned in a comment above: we have two dates so we can make this association with a boolean value which is a log neater (no risk for a typo in a string for example). So for example, instead of name use isFromDate: boolean. then this if/else turns into if(isFromDate) {} else {} without a costly and error prone comparison.

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+    const isModified = this.fromField() !== this.#defaultFromDate();
+    if (isModified !== wasModified) {
+      this.isModified.set(isModified);
+      this.edited.emit(['access_ends', isModified]);
+    }
+  };
+
+  cancel = () => {
+    if (this.isModified()) {
+      this.fromField.update(() => this.#defaultFromDate());
```
SilverLinings89, line 301: I think you should also reset the other field, no?

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+      this.edited.emit(['access_ends', false]);
+    }
+    this.isOpen.set(false);
+  };
+
+  save = () => {
+    this.onDateSelected(new Date(this.fromField() as Date), 'access_starts');
+    this.onDateSelected(new Date(this.untilField() as Date), 'access_ends');
+    if (!this.saveDisabled() && this.fromField() && this.untilField()) {
```
SilverLinings89, line 310: If the latter two were undefined, ts execution already broke in the two lines above.

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
```diff
@@ -0,0 +1,327 @@
...
+      const untilDate = new Date(this.untilField()!);
+      if (fromDate && untilDate) {
+        const fromISO = fromDate.toISOString();
+        const untilISO = untilDate.toISOString();
+        if (this.isModified()) {
+          const saveMap = new Map<keyof AccessRequest, string>();
+          saveMap.set('access_starts', fromISO);
+          saveMap.set('access_ends', untilISO);
+          this.saved.emit(saveMap);
+          this.edited.emit(['access_ends', false]);
```
SilverLinings89, line 321: I don't really understand why you do this...

src/app/shared/services/config.service.ts
```diff
@@ -180,6 +181,14 @@ export class ConfigService {
     return this.#config.access_grant_max_days;
   }
 
+  /**
+   * Gets the maximum factor of max days a data steward can extend an access grant
+   * @returns the number of days
+   */
+  get accessGrantMaxExtend(): number {
```
SilverLinings89, line 188: Should these access grant max extend changes be here? I don't see this value used anywhere and access grants arent discussed in this PR, right?

src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.ts
ac-jorellanaf: The emitting of the edited? Because that lets the dialog component know whether it should show the warning about unsaved changes, and when saved, it should also make sure the dialog component doesn't show that warning.

ac-jorellanaf:
> There is a mis-alignment between for the date editing: image
> 
> This way, it looks like it only edits the first value, not both.

The issue is that we can't really do it so that it's in the middle unless we refactor all the table-based elements into more flexible grids. This is exceedingly out of scope for this PR though.

src/app/access-requests/features/access-request-dialog/access-request-dialog.component.ts
```diff
@@ -244,12 +250,12 @@ export class AccessRequestDialogComponent {
...
   updateUntilRangeForFromValue(date: Date): void {
-    const newUntilMinDate = new Date(
-      date.getTime() + this.#config.accessGrantMinDays * MILLISECONDS_PER_DAY,
-    );
-    const newUntilMaxDate = new Date(
-      date.getTime() + this.#config.accessGrantMaxDays * MILLISECONDS_PER_DAY,
-    );
+    const newUntilMinDate = new Date(date);
+    newUntilMinDate.setHours(23, 59, 59, 999);
+    newUntilMinDate.setDate(date.getDate() + this.#config.accessGrantMinDays);
```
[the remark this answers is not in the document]
SilverLinings89: no. What this does is it does the action of the second and first line in one (by using a constructor that sets all the date and time values. Date objects allow setting the date (year, month, day) and time (hour etc.) independently. What this does is it first sets both date (a dummy) and time (as specified) and then only overwrites the date part in the second line. The previous version first sets everything to now and then overwrites first the time and then the date. That's one operation more. You could also first determine the year, month and day values individually from date and then just do one constructor call with all the values.

ac-jorellanaf:
> > > There is a mis-alignment between for the date editing: image
> > > This way, it looks like it only edits the first value, not both.
> > 
> > 
> > The issue is that we can't really do it so that it's in the middle unless we refactor all the table-based elements into more flexible grids. This is exceedingly out of scope for this PR though.
> 
> You could do it with a rowspan 2 for the date area but it's not that important.

I tried very hard to make it work but it, for some reason, doing it this way makes the second cell with the label be pushed down. It also doesn't vertically align the contents, and `display:flex` breaks the rowspan. Having a flexbox inside the row also does not work because it does not use the entire height of the 2 rowspan cell either unless the cell is a flexbox but then it also breaks rowspan so it does not work. 😓

PR #154, opened 2025-06-17
[DRAFT] Transpiler UI Demo Implementation

SilverLinings89:
This is to track progress in this direction. 
This is my current state that I think has all the required features and is fully integrated into the angular app as a version of how we could do this. Looking for feedback.

### DIFF
--- a/src/app/app.routes.ts
+++ b/src/app/app.routes.ts
@@ -37,6 +37,14 @@ export const routes: Routes = [
     title: 'Validate Metadata',
   },
   {
+    path: 'schemapack-playground',
+    loadComponent: () =>
+      import(
+        './tools/features/schemapack-playground/schemapack-playground.component'
+      ).then((m) => m.SchemapackPlaygroundComponent),
+    title: 'Schemapack Playground',
+  },
+  {
     path: 'dataset/:id',
     loadComponent: () =>
       import('./metadata/features/dataset-details/dataset-details.component').then(
--- a/src/app/portal/features/site-header-nav-buttons/site-header-nav-buttons.component.html
+++ b/src/app/portal/features/site-header-nav-buttons/site-header-nav-buttons.component.html
@@ -15,11 +15,28 @@
 >
 <a
   mat-button
-  routerLink="/metadata-validator"
-  class="mx-1"
-  data-umami-event="Site Header Browse Clicked"
-  >Tools</a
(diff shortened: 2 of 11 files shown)


SilverLinings89:
Just as an additional bit of information: I thought we could introduce a category tools in the main menu since we will also be launching Schemapack to the public soon and we could recommend it there and reference some docs and the repo there.

Cito:
Great work and good idea, we could also include the CLI docs there. Will review this after my vacation. If we include this, we will also need to add some e2e testing and automated Python formatting/linting/type checking.

Cito (approved):
I reviewed this now in more detail, played with it to get a feeling for how it works, and pushed some small improvements.

Overall, I like this approach to offer some of our CLIs inside the Data Portal as well. The PoC  shows that this works very well and fast enough in practice, can be nicely integrated with the rest of the Data Portal, and is pretty easy to maintain once you understand how it works.

This approach also keeps the footprint in our bundles and source code very small - I removed Pyodide from the bundled packages, so that everything is only loaded via CDN from the Internet on demand.

I also like the UI for this tool a lot, very functional and clear. The stepper looks even nicer than the one included in Material. The validation errors could be formatted more nicely as HTML with working links etc. instead of just being shown as raw text, which would then probably look even nicer than the output in the CLI.

Some ideas for things we should do to get this from PoC to production:
- Add some Python tooling to our dev container, so that Python can be formatted and linted, maybe also type checked and tested.
- We should create a reusable service and wrapper with well-tested and well-defined API. that allows running any Python CLI from JavaScript.
- This means if we don't use Pyodide for anything else, the amount of custom Python code may be very small and static, so we might actually not even need additional Python tooling and testing in the dev container.
- The CLI tools could also be improved to provide programmatic interfaces instead of just running on the command line, then this might be even easier to use from Pyodide. They could also allow sending output to stdout or strings so that we can work purely in memory.
- We should move the generalized service that runs any CLIs to /shared/services.
- The tools could then go to the corresponding contexts (vertical slices), i.e. `/metadata` in this case, instead of `/tools`.
- The Tools menu entry should open a submenu with all the offered tools. Some of them could also be just JavaScript.
- We should simplify the excessive debug logging and error catching. If there is an error, it should fail and produce the output it produces anyway (from Python or the CLI).
- The stepper should be made reusable as well.
- Nicer output for the validation errors as mentioned above.

PR #155, opened 2025-06-24
Properly transform and validate ticket IDs

Cito:


src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
```diff
@@ -71,24 +82,33 @@ export class AccessRequestFieldEditComponent implements OnInit {
...
+   * and set the proper error message if the field is invalid.
+   * @param control - the form control for the field
+   */
+  #handleControl = (control: FormControl) => {
+    if (this.name() === 'ticket_id') {
+      // if the field is prefixed with (parts of) the base ticket URL, remove that prefix
+      const baseUrl = this.#baseTicketUrl;
+      let value = this.field() || '';
+      const i = value.lastIndexOf('/');
+      if (baseUrl && i >= 0 && baseUrl.endsWith(value.substring(0, i + 1))) {
```
ac-jorellanaf, line 100:
```suggestion
      if (baseUrl && i >= 0 && baseUrl.startsWith(value.substring(0, i + 1))) {
```
Does not work with endsWith for me.
Also should we not strip the URL when we save rather than when the user pastes it? Otherwise it is a little confusing, but it's not a big deal.

src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
```diff
@@ -71,24 +82,33 @@ export class AccessRequestFieldEditComponent implements OnInit {
...
+        value = value.substring(i + 1);
+        control.setValue(value);
+      }
+      this.validationError.set(control.invalid ? ERROR_TICKET_ID : '');
+    }
+  };
+
+  changed = (control?: FormControl) => {
```
ac-jorellanaf, line 108: By the way, and this is true also for the date editing, if you click cancel after you have already saved a previous edit, the data goes back to what was originally there before the first edit, rather than the value saved after the first edit.

ac-jorellanaf (changes requested):
Everything works well besides one thing.

src/app/access-requests/features/access-request-field-edit/access-request-field-edit.component.ts
ac-jorellanaf: This might have to do with the way mocking works, but I have not yet explored it thoroughly.

PR #157, opened 2025-06-24
Change display of access duration dates in manager dialog

ac-jorellanaf:
- Made it so the access duration is displayed in a single line in the access request manager dialog when not editing
- Improved handling of UTC dates by duration editor and request access dialog; now all from dates should be set to 00:00:00:0 UTC and all end dates to 23:59:59:999 UTC of the same day as the selected date (not the case before)
- Minor improvement to mock data


src/app/access-requests/features/access-request-duration-edit/access-request-duration-edit.component.html
```diff
@@ -1,105 +1,82 @@
...
-          class="smaller-chip interactive-chip save-edit-button mb-5"
-          (click)="save()"
-          [disabled]="saveDisabled()"
-          ><mat-icon>check</mat-icon></mat-chip
-        >
-      } @else {
+  @if (isOpen()) {
+    <tr>
+      <th class="w-36 min-w-36 pb-5 text-start">Access duration:</th>
+      <td class="min-w-60">
```
Cito, line 5:
```suggestion
      <td class="min-w-60 pb-1">
```

otherwise the error message is too close to the next row

Cito (approved):
Looks much better now. Just two very small suggestions.

Cito (approved):

PR #162, opened 2025-07-08
Add date pipe that supports IANA time zone names

Cito:
The DatePipe that is shipped with Angular does not support IANA time zones, only fixed zone offsets. We need dynamic zones for the access and end dates. Therefore, I added an improved date pipe that supports the dynamic IANA zones.

### DIFF
--- a/src/app/shared/pipes/date.pipe.ts
+++ b/src/app/shared/pipes/date.pipe.ts
@@ -1,0 +1,85 @@
+/**
+ * Improved DatePipe that supports IANA time zone names.
+ * @copyright The GHGA Authors
+ * @license Apache-2.0
+ */
+
+import { DatePipe as CommonDatePipe } from '@angular/common';
+import { inject, Pipe, PipeTransform } from '@angular/core';
+
+/**
+ * This is an improved DatePipe that also supports IANA time zone names.
+ *
+ * The existing Angular DatePipe currently only supports time zone offsets.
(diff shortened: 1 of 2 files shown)


src/app/shared/pipes/date.pipe.ts
```diff
@@ -0,0 +1,86 @@
...
+ * the use of IANA time zone names like 'Europe/Berlin'.
+ *
+ * Please use either time zone offsets or IANA time zone names, but do not
+ * use time zone abbreviations such as 'CET', as they can be ambiguous and
+ * lead to incorrect results depending on the runtime environment.
+ *
+ * We assume the Intl API is supported since all modern browsers have it.
+ * If it is not supported, it behaves like the original DatePipe.
+ */
+@Pipe({ name: 'DatePipe' })
```
ac-jorellanaf, line 26: Since we need to alter the imports if we want to use the pipe anyway, why not use a name that does not clash with the angular data pipe?

ac-jorellanaf (approved):
LGTM

src/app/shared/pipes/date.pipe.ts
Cito: The idea was that this is actually a (temporary) replacement for the Angular pipe until it is fixed, so the templates do not need to be changed. I will consider renaming it in the next PR where I will actually use it.

PR #170, opened 2025-07-22
Added skip link to data portal (GSI-1768)

ac-jorellanaf:


src/app/app.component.html
```diff
@@ -1,10 +1,17 @@
 <div class="flex min-h-screen flex-col">
+  <div class="flex w-full justify-center">
+    <a
```
Cito, line 3: You should add a short comment explaining that this is an a11y feature, because the next dev will wonder what it's supposed to do.

Cito (changes requested):

PR #171, opened 2025-07-23
Add user management list view (GSI-1734)

Cito:
This PR implements the main list view for the user management page.

It already includes a stub for the detail page and the navigation between the two.

The corresponding view transition will be implemented in a separate PR since it turned out to be not so easy as anticipated.

The filter section will also be implemented in a separate PR.

### DIFF
--- a/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
+++ b/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
@@ -9,7 +9,7 @@
       >
         <mat-icon>chevron_left</mat-icon>
       </button>
-      <h1 class="text-3xl font-bold">User: {{ currentUser.displayName }}</h1>
+      <h1>User: {{ currentUser.displayName }}</h1>
     </div>
 
     <div class="rounded-lg bg-white p-6 shadow-md">
--- a/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
+++ b/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
@@ -1,5 +1,5 @@
 @if (user(); as currentUser) {
-  <div class="container mx-auto p-6">
+  <div class="mx-auto max-w-[1200px] p-6">
     <div class="clearfix mb-6">
       <button
         mat-icon-button
--- a/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
+++ b/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
@@ -53,14 +53,14 @@
                 >Registration Date</span
               >
               <p class="mt-1 text-sm text-gray-900">
-                {{ currentUser.registration_date | date }}
+                {{ currentUser.registration_date | date: friendlyDateFormat }}
               </p>
             </div>
             <div>
               <span class="block text-sm font-medium text-gray-700">Roles</span>
               <div class="mt-1 flex flex-wrap gap-1">
-                @for (role of currentUser.roles; track role) {
-                  <mat-chip>{{ role }}</mat-chip>
+                @for (roleName of currentUser.roleNames; track roleName) {
+                  <mat-chip>{{ roleName }}</mat-chip>
                 } @empty {
                   <p class="text-sm text-gray-500">No roles assigned</p>
                 }
--- a/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
+++ b/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
@@ -1,15 +1,15 @@
 @if (user(); as currentUser) {
   <div class="container mx-auto p-6">
-    <div class="mb-6 flex items-center gap-4">
+    <div class="clearfix mb-6">
       <button
         mat-icon-button
         (click)="goBack()"
         aria-label="Go back to user list"
-        class="text-primary"
+        class="text-primary float-left mt-1 mr-4"
       >
-        <mat-icon>arrow_back</mat-icon>
+        <mat-icon>chevron_left</mat-icon>
       </button>
-      <h1 class="text-3xl font-bold">User: {{ userDisplayName() }}</h1>
+      <h1 class="text-3xl font-bold">User: {{ currentUser.displayName }}</h1>
     </div>
 
     <div class="rounded-lg bg-white p-6 shadow-md">
--- a/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
+++ b/src/app/auth/features/user-manager-detail/user-manager-detail.component.html
@@ -1,0 +1,96 @@
+@if (user(); as currentUser) {
+  <div class="container mx-auto p-6">
+    <div class="mb-6 flex items-center gap-4">
+      <button
+        mat-icon-button
+        (click)="goBack()"
+        aria-label="Go back to user list"
+        class="text-primary"
+      >
+        <mat-icon>arrow_back</mat-icon>
+      </button>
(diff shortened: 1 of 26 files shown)


src/app/auth/features/user-manager-detail/user-manager-detail.component.html
```diff
@@ -0,0 +1,96 @@
+@if (user(); as currentUser) {
+  <div class="container mx-auto p-6">
```
ac-jorellanaf, line 2:
```suggestion
  <div class="mx-auto max-w-[1200px] p-6">
```

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
```diff
@@ -0,0 +1,37 @@
...
+:host {
+  display: block;
+  padding-bottom: 1em;
+  background-color: var(--mat-sys-surface-container-low);
+}
+
+.container {
+  max-width: 1200px;
+}
+
```
ac-jorellanaf, line 18:
```suggestion
```

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+@use '@angular/material' as mat;
+
+:host {
+  display: block;
+  padding-bottom: 1em;
+  background-color: var(--mat-sys-surface-container-low);
+}
+
```
ac-jorellanaf, line 14: Why not simply have a div wrapping the if and else and give it these classes? It's the same effect and avoids having to use `:host`.

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
```diff
@@ -0,0 +1,37 @@
...
+ * @copyright The GHGA Authors
+ * @license Apache-2.0
+ */
+
+@use '@angular/material' as mat;
+
+:host {
+  display: block;
+  padding-bottom: 1em;
+  background-color: var(--mat-sys-surface-container-low);
```
ac-jorellanaf, line 12: Not sure I like the separate background colour. Maybe make the borders rounded at least, and a bit more contrasting?

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Status color overrides if needed
+.status-active {
+  color: var(--mat-sys-primary);
+}
+
+.status-inactive {
+  color: var(--mat-sys-error);
+}
```
ac-jorellanaf, line 37: We don't need these. Tested 🙂

src/app/auth/features/user-manager-list/user-manager-list.component.html
```diff
@@ -0,0 +1,109 @@
...
+        <th mat-header-cell *matHeaderCellDef class="min-w-48" mat-sort-header>
+          Email
+        </th>
+        <td mat-cell *matCellDef="let user">
+          {{ user.email }}
+        </td>
+      </ng-container>
+
+      <ng-container matColumnDef="name">
+        <th mat-header-cell *matHeaderCellDef class="min-w-40" mat-sort-header>
```
ac-jorellanaf, line 23:
```suggestion
        <th mat-header-cell *matHeaderCellDef class="min-w-48" mat-sort-header>
```
We should avoid having names wrap

src/app/auth/features/user-manager-list/user-manager-list.component.html
```diff
@@ -0,0 +1,109 @@
...
+        </td>
+      </ng-container>
+
+      <tr mat-header-row *matHeaderRowDef="columns"></tr>
+      <tr
+        mat-row
+        *matRowDef="let row; columns: columns"
+        class="duration-200 hover:cursor-pointer hover:bg-black/[0.05] hover:transition"
+        (click)="openDetails(row)"
+      ></tr>
```
ac-jorellanaf, line 87: We should add some highlighting on hover, much like the other management tables, but this can be in a separate PR.

src/app/auth/features/user-manager-detail/user-manager-detail.component.html
```diff
@@ -0,0 +1,96 @@
...
+    <div class="clearfix mb-6">
+      <button
+        mat-icon-button
+        (click)="goBack()"
+        aria-label="Go back to user list"
+        class="text-primary float-left mt-1 mr-4"
+      >
+        <mat-icon>chevron_left</mat-icon>
+      </button>
+      <h1 class="text-3xl font-bold">User: {{ currentUser.displayName }}</h1>
```
ac-jorellanaf, line 12:
```suggestion
      <h1 class="font-bold">User: {{ currentUser.displayName }}</h1>
```
I think this is redundant

src/app/auth/features/user-manager-detail/user-manager-detail.component.html
```diff
@@ -0,0 +1,96 @@
...
+    </div>
+
+    <div class="rounded-lg bg-white p-6 shadow-md">
+      <div class="grid grid-cols-1 gap-6 md:grid-cols-2">
+        <div>
+          <h2 class="mb-4 text-xl font-semibold">Basic Information</h2>
+          <div class="space-y-3">
+            <div>
+              <span class="block text-sm font-medium text-gray-700">Name</span>
```
ac-jorellanaf, line 21:
```suggestion
              <span class="block text-sm font-bold text-gray-700">Name</span>
```
The headers should be bolded for a better look. It's a bit too hard to parse as it stands right now

src/app/auth/features/user-manager-detail/user-manager-detail.component.html
```diff
@@ -0,0 +1,96 @@
+@if (user(); as currentUser) {
+  <div class="container mx-auto p-6">
+    <div class="clearfix mb-6">
+      <button
+        mat-icon-button
+        (click)="goBack()"
+        aria-label="Go back to user list"
+        class="text-primary float-left mt-1 mr-4"
```
ac-jorellanaf, line 8:
```suggestion
        class="text-primary md:float-left mt-1 mr-4"
```
For a better look on small screens

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
Cito: Removed this, since the detail page in this page is only meant as a paceholder and starting point anyway.

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
Cito: See above.

src/app/auth/features/user-manager-detail/user-manager-detail.component.scss
Cito: Removed.

src/app/auth/features/user-manager-list/user-manager-list.component.html
Cito: I have added a separate task GSI-1772 to make the look and feel consistent across all admin modules.

src/app/auth/features/user-manager-detail/user-manager-detail.component.html
Cito: I'm leaving this as is since this needs to be redesigned according to our wireframes anyway.

PR #172, opened 2025-07-23
Minor responsiveness improvements

ac-jorellanaf:
Header, navbar, and management pages.

Fixed issues:
- Navbar title too wide, overlaps with menu button
- Cannot navigate to submenus without sidenav hiding
- Main page should not have justified text for small screens
- IVA and AR manager filter not responsive


Cito (changes requested):
Looks good, I like the toggle for the filters.

Main suggestion is to improve the sidebar nav closing, see below.

PR #173, opened 2025-07-24
Update packages and add MCP server config

Cito:


ac-jorellanaf (dismissed):
Works well, but I don't think I see any server config changes?

ac-jorellanaf (approved):

PR #178, opened 2025-07-31
Adapt access request manager to use a detail view (GSI-1744)

Cito:
This PR adapts the access request manager so that it uses a detail page with transition instead of a dialog.

The detail page tries to fetch the access request from the list that has been fetched in the list view already. However, if this fails, it also tries to fetch it directly. This is necessary if this page is linked to from another page, like the access or user manager.

### DIFF
--- a/src/app/access-requests/features/access-request-manager-detail/access-request-manager-detail.component.spec.ts
+++ b/src/app/access-requests/features/access-request-manager-detail/access-request-manager-detail.component.spec.ts
@@ -130,8 +130,8 @@ describe('AccessRequestManagerDetailComponent', () => {
     expect(button).toBeEnabled();
   });
 
-  it('should have an enabled close button', () => {
-    const button = screen.getByRole('button', { name: 'Close' });
+  it('should have an enabled back button', () => {
+    const button = screen.getByRole('button', { name: 'Back' });
     expect(button).toBeEnabled();
   });
 });


src/app/access-requests/features/access-request-manager-detail/access-request-manager-detail.component.html
```diff
@@ -233,65 +248,79 @@ <h2 mat-dialog-title>Access Request Detail</h2>
...
-          autoFocus
           mat-stroked-button
-          class="w-full sm:w-auto"
-          (click)="cancel()"
-          data-umami-event="Access Request Manager Cancel Clicked"
+          (click)="goBack()"
+          data-umami-event="Access Request Manager Close Clicked"
         >
-          Cancel
+          Close
```
ac-jorellanaf, line 259:
```suggestion
          Cancel
```

src/app/access-requests/features/access-request-manager-detail/access-request-manager-detail.component.html
```diff
@@ -233,65 +248,79 @@ <h2 mat-dialog-title>Access Request Detail</h2>
...
+            class="w-full sm:w-auto"
+            (click)="goBack()"
+            data-umami-event="Access Request Manager Close Clicked"
           >
-            @if (data.status === 'allowed') {
-              Revoke access
-            } @else {
-              Deny
-            }
+            Close
```
ac-jorellanaf, line 270:
```suggestion
            Cancel
```

ac-jorellanaf (changes requested):
Two issues here.
First, the applied filter does not persist when navigating back to the access request list. It resets every time.
Second, the caching does seem to work (at least using mocks) when navigating from the list, but do a hard reload or navigate to a details page outside the list page, and the UI does two calls: one for the current access request, and one for _all_ access requests. This is undesired behaviour.

src/app/access-requests/features/access-request-manager-detail/access-request-manager-detail.component.html
Cito: I changed this to "Close" because the page makes the changes immediately (like when you select an IVA or click on of the green checkmarks), while a "Cancel" would somewhat imply that these changes are rolled back.

Maybe we should label it "Cancel" only if one of the field editors is open and has a pending change in its input field, and label it "Close" otherwise? Another option would be to add a navigation guard that would show a confirmation dialog before closing the page and navigating away in this state. I think one of these two measures would be sufficient. Which one do you like better?

ac-jorellanaf: It really is not a 'closing' action though. Maybe 'return' or 'back' would be better?

Cito: Particularly if we make it an actual back button, then "Back" would be a better than "Close". Should we do that?

Regarding pending changes, I still have the question whether we should add a guard or just rename the button to "Cancel" while they are pending.

src/app/access-requests/features/access-request-manager-detail/access-request-manager-detail.component.html
Cito: Already handled in the conversion above.

Cito:
Just for the records:

Filter issue: It turned out that the filter reset only happens in one edge case, when there is only a pending status filter, and that is removed. Going back applied the pending filter again. This was due to a mismatch between the default filters in the getters and setters for the filters. Fixed that for the other services as well.

Fetching all access requests issues: The problem was that the access request service always fetches the requests for the logged in user on every page, because they are used in some parts of the application. It is independent of the access request detail view though and not triggered by it.

ac-jorellanaf (approved):

PR #185, opened 2025-08-06
Improve local cache updates

Cito:
This PR improves the methods that update resources locally where data is changed in the backend, so that no reload is required.

We now only update resources that aren't in an error state, because otherwise the update will raise an error. We now also update the new resources that fetch single objects instead of an array of objects.

### DIFF
--- a/src/app/verification-addresses/services/iva.service.ts
+++ b/src/app/verification-addresses/services/iva.service.ts
@@ -125,6 +125,7 @@ export class IvaService {
    * Signal that gets all IVAs filtered by the current filter
    */
   allIvasFiltered = computed(() => {
+    if (this.allIvas.error()) return [];
     let ivas = this.allIvas.value();
     const filter = this.#allIvasFilter();
     if (ivas.length && filter) {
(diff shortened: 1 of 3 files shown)


src/app/verification-addresses/services/iva.service.ts
```diff
@@ -125,6 +125,7 @@ export class IvaService {
    * Signal that gets all IVAs filtered by the current filter
    */
   allIvasFiltered = computed(() => {
+    if (this.allIvas.error()) return [];
```
SilverLinings89, line 128: I like early returns a lot for this situation because it reduces the complexity in the code below. In the situations where it works (only one error state), we should make it a style preference for the project.

SilverLinings89 (approved):
LGTM

Cito:
I agree this should be refactored a bit to reduce complexity and duplication. More generally I think we should standardize our patterns how we use httpResources now that the code has grown a lot and the Angular support has become more stable.

Will merge as is for now. Would be good to just refactor and improve a11y in the next release cycle.

PR #186, opened 2025-08-06
Add link to user and filter by email (GSI-1745)

Cito:
Because the IVA manager does not show much information about the user, I added a link to the user detail page.

Also allowed searching for email addresses with the name field, like in the other editors.

We can consider showing the email in parens after the user name, but then the table would probably look too growded and overflow.


SilverLinings89 (approved):
no comments :)

PR #197, opened 2025-08-26
Tables changed to div in global summary

ac-jorellanaf:


ac-jorellanaf:
> I don't really see the point of this. I wrote this in the accessibility doc comment response, too. The point isn't that tables are somehow bad. In the early days of the web standards when styling was very difficult, many web devs fell back to tables for page structure (to for example fix a menu on the left to the left. When people write guides about accessibility these days, that pattern is meant when someone says that you shouldn't use tables for page structure. These views logically use tables. In this circumstance, tables are the semantically correct thing to use. (see for example here for context: https://thehistoryoftheweb.com/tables-layout-absurd/)

From W3 (https://www.w3.org/WAI/tutorials/tables/):

> This tutorial provides guidance for creating tables used to display data in a grid. This tutorial does not apply to tables used for layout. **As a general rule, tables aren’t meant to be used for layout purposes. Instead, a best practice is to use Cascading Style Sheets (CSS) for visual presentation.**

A table without th is also considered as failing the test for Technique H51 in WCAG 2.1 (https://www.w3.org/WAI/WCAG21/Techniques/html/H51#tests).

Silktide also highlights it as a problem.

And please take a look at my reply in the other review.

ac-jorellanaf:
> I had seen that and I agree that it can be seen as an issue that we aren't using headers but if, for example, we added headers, like "count" and "platform" for experiments, "sex" and "count" for individuals and "count" and "filetype" that would actually make the page easier to understand.

This is fine, and I agree, but as it was without headers, this _is_ a layout table. Not to split hairs or anything, but it is _required_ that all <table> elements have at least one header row/column.
Screen reader-wise, there was no major issue with leaving them as divs as they were, since they are just read one after the other, and this is _identical_ to them being inside a table with headers, the only difference is that they get the header information before the number and platform, etc.
Nevertheless, I added headers to the tables since it does make sense to add more context rather than not, but I just want to clarify that the tables, regardless of what kind of thing they contain **_are_ layout tables when they do not have headers** and therefore not compliant with WCAG 2.1 AA, and the only difference for screen readers (as far as I can tell) between the table and the div as it was before, is the announcement of the headers themselves,

Cito (commented):
I agree with Pascal that this is actually tabular data, and the table is not just a layout table. I also agree with Jordy that in this case it should contain a table header. 

The case here is a bit difficult because we want this is the front page and we want to look stylish, not boring, like an infographics, and not too crowed. Headers somehow destroy this nice look, even if they are informative for screen readers.

What about using a technique like [this](https://www.davidmacd.com/blog/make-a-table-header-row-invisible.html) to hide the thead in the visualy layout, as a compromise?

Cito (changes requested):
This works for me, but I suggest to change the class name to visually-hidden because visibly-hidden sound as if it would set visibility:hidden which is exactly what it should not do.

Also, even if this works for our use case in chrome and firefox, I would add some more styles, like shown here: https://blog.logrocket.com/design-accessibility-css-visually-hidden-class/

Particularly, border, margin and padding should be set to 0.

Also, please add a comment to the style sheet explaining the purpose of the class.

src/styles.scss
```diff
@@ -265,6 +265,16 @@ button.error {
   );
 }
 
+// Class for hiding visually hiding elements but allowing them to remain visible to screen readers
```
Cito, line 268:
```suggestion
// Class for visually hiding elements while keeping them visible in screen readers
```

duplicate word

Cito (changes requested):

PR #199, opened 2025-08-28
Header and footer navigation links now much more accessible

ac-jorellanaf:
Also fixed accessibility for all icons in the website

### DIFF
--- a/src/app/access-requests/features/access-grant-details/access-grant-manager-details.component.html
+++ b/src/app/access-requests/features/access-grant-details/access-grant-manager-details.component.html
@@ -60,16 +60,9 @@
           <mat-card-content class="min-w-2xl md:min-w-xl">
             <p>
               <strong class="inline-block w-44">Email:</strong>
-              <<<<<<< HEAD
-              <a [href]="'mailto:' + ag.user_email" target="_blank" tabindex="-1"
-                >{{ ag.user_email
-                }}<mat-icon aria-hidden="true" fontIcon="open_in_new"></mat-icon>
-              </a>
-              =======
               <a appExtLink [href]="'mailto:' + ag.user_email" tabindex="-1">{{
                 ag.user_email
               }}</a>
-              >>>>>>> origin/main
             </p>
             <p>
               <strong class="inline-block w-44">Name:</strong>


src/app/access-requests/features/access-grant-details/access-grant-manager-details.component.html
```diff
@@ -11,7 +11,7 @@
         aria-label="Go back to access grant list"
         class="text-primary float-left mt-1 mr-4"
       >
-        <mat-icon>chevron_left</mat-icon>
+        <mat-icon fontIcon="chevron_left"></mat-icon>
```
Cito, line 14: Is there a reason why you changed that from the ligature approach? In how far does it help with a11y?

Cito: These template changes are hard to review. It would be helpful if you summarize in the description of the PRs what you did (e.g. changed mat-icons to use fontIcon, added aria-labels, moved navigation elements to lists) and why (if not obvious).

If there is a recommeded way to use font icons, it would be good to have this mentioned in our a11y docs as well.

ac-jorellanaf: Yes the SilkTide one, but also looked at the A11y tree in Firefox and Chrome.

Cito: I can confirm that when there is an aria-label, it is shown instead of the icon name in the SilkTide reader, regardless of whether I use fontIcon or literal icon name. 

So are you saying this is solved and we can continue to use literal icon names? In that case I would prefer keeping the literal icon names, it looks cleaner and simpler, and works as a fallback if we forget to set an aria-label.

ac-jorellanaf: We should _not_ forget the aria-label though. 😄
The SilkTide screen reader says 'open underscore in underscore new', which is very hard to parse, in my opinion. It doesn't really matter anyways. As I said, this only happened with the literal icons _and_ having them float in the element.

I think the ideal would be 1, since the second one ends up being read as two different elements by screen readers.

Cito: Yes, underscored names are ugly in this case.

I fixed the syntax for options 2, the mat-icon should be hidden to screen readers (aria-hidden is the default for mat-icon and with fontIcon=... it should work). Option 1 is simpler, but as you say we should then not forget to add the label.

Cito: In option 2 the mat-icon and span should be siblings (tried to indicate this with the self-closing tag).

Cito: Just for the records, we decided to consistently use the fontIcon attribute, because most of the icons are purely decorative, and screen readers tend to not show it, contrary to the ligature name as inline text. We also decided to add an Angular directive for the `open_in_new`  case, that would add a target, aria-label and mat-icon to the link. That way, the information that this is opened in a new tab would be shown in the link, but not in an enclosing paragraph where it would be only disturbing.

src/app/access-requests/features/access-grant-details/access-grant-manager-details.component.html
```diff
@@ -60,9 +60,16 @@ <h1>Access Grant Detail</h1>
...
+              <<<<<<< HEAD
+              <a [href]="'mailto:' + ag.user_email" target="_blank" tabindex="-1"
+                >{{ ag.user_email
+                }}<mat-icon aria-hidden="true" fontIcon="open_in_new"></mat-icon>
+              </a>
+              =======
               <a appExtLink [href]="'mailto:' + ag.user_email" tabindex="-1">{{
                 ag.user_email
               }}</a>
+              >>>>>>> origin/main
```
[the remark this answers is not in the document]
ac-jorellanaf: Very odd. Must not have saved after the resolving. Fixed now

PR #202, opened 2025-09-04
Added missing tabindex rule to ESLint

ac-jorellanaf:
All other a11y rules already enabled
Made it so ESLint also checks inline templates (if any)


Cito (approved):
Makes sense.

I wonder why it's not in the accessibility.rules. Should we create an issue/PR in the angular-eslint repo?

PR #215, opened 2025-10-07
Several improvements to access grant details page (GSI-1893)

ac-jorellanaf:
- Added more details to IVA section in page (required an API call for all access requests)
- Added logging for multiple access requests (currently all access request log events are shown above any access grant log events, even if the former happened after the latter, this could be improved, however)
-Improved warning for multiple access reque sts


Cito (changes requested):
Minor simplification.

PR #294, opened 2026-04-13
Use new syntax for eslint-plugin-boundaries (GSI-2243)

Cito:
We use [eslint-plugin-boundaries](https://github.com/javierbrea/eslint-plugin-boundaries) to define the rules for importing in our architecture matrix. The latest version 6 introduces a new object-based syntax for these rules, instead of strings and tuples. This PR refactors the rules to use the new syntax.


eslint.config.js
```diff
@@ -81,192 +81,209 @@ export default [
...
+                ],
+              },
             },
             // main may only import config and main app modules
             {
-              from: ['main'],
-              disallow: ['*'],
+              from: { type: 'main' },
+              disallow: { to: { type: '*' } },
               message: 'Main modules may only import config and main app',
```
maryam-sheikhi: Should the meassage update to : 'Main modules may only import config, main app and mock'? because mock is also allowed now.

maryam-sheikhi (dismissed):
The migration is thorough, and correct, I just added a comment.

eslint.config.js
Cito: Ok, changed.

PR #299, opened 2026-04-17
Upload token creation (GSI-2095)

Cito:
This PR adds a dialog to create upload tokens to the user account page.


src/app/upload/features/user-upload-grants-list/user-upload-grants-list.ts
```diff
@@ -58,11 +69,27 @@ export class UserUploadGrantsListComponent {
   protected submittingBoxId = signal<string | null>(null);
 
   /**
-   * Placeholder handler for the "Create token" button.
-   * Token creation will be implemented in a follow-up PR.
+   * On initialisation, load the current user's IVAs.
    */
-  createToken(): void {
-    this.#notification.showInfo('Upload token creation is not yet implemented.');
+  ngOnInit(): void {
```
maryam-sheikhi: I think it is better to check  if there are open grants first, and then loading IVAs, to improve performance. Ex:  if (this.openGrants().length > 0) {
      this.#iva.loadUserIvas();
    }

maryam-sheikhi (commented):
The changes are well implemented, I only added a comment for performance improvement.

src/app/upload/features/user-upload-grants-list/user-upload-grants-list.ts
Cito: Good idea, changed in 4e212db. Needs and effect though since ngOnInit() runs only once.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
```diff
@@ -0,0 +1,160 @@
...
+  protected iva = this.grant.iva;
+  protected readonly friendlyDateFormat = FRIENDLY_DATE_FORMAT;
+
+  protected model = signal({ pubkey: '' });
+  protected pubkeyField = viewChild.required(PubkeyFieldComponent);
+  protected uploadForm = form(this.model, (p) => {
+    apply(p.pubkey, PubkeyFieldComponent.schema);
+  });
+
+  token = signal('');
```
maryam-sheikhi: I think it is better to set token signals as protected.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
```diff
@@ -0,0 +1,160 @@
...
+    if (detail) dialogMsg += `: ${detail}`;
+    dialogMsg += '.';
+    this.tokenError.set(dialogMsg);
+
+    let notificationMsg = 'Cannot create upload token';
+    if (detail) notificationMsg += `: ${detail}`;
+    notificationMsg += '!';
+    this.#notify.showError(notificationMsg);
+
+    console.debug(error);
```
maryam-sheikhi, line 118: Do we need console.debug(error) in production?

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
```diff
@@ -0,0 +1,160 @@
...
+    const workPackage: UploadWorkPackageRequest = {
+      type: 'upload',
+      research_data_upload_box_id: this.grant.box_id,
+      user_public_crypt4gh_key: this.pubkeyField().trimmedKey,
+    };
+
+    this.resetToken();
+    this.tokenIsLoading.set(true);
+
+    this.#wpService.createWorkPackage(workPackage).subscribe({
```
maryam-sheikhi, line 136: Do you unsubscribe later?

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
```diff
@@ -0,0 +1,160 @@
...
+    });
+  }
+
+  /**
+   * Clear current token to allow creating a new one.
+   */
+  resetToken(): void {
+    this.token.set('');
+    this.tokenIsLoading.set(false);
+    this.tokenError.set('');
```
maryam-sheikhi, line 152: Maybe we should also reset tokenExpiration, Ex: Add  this.tokenExpiration.set('');

eslint.config.js
```diff
@@ -148,6 +148,22 @@ export default [
...
+                  { type: 'model', captured: { context: 'ivas' } },
+                  { type: 'pipe', captured: { context: 'ivas' } },
+                  { type: 'model', captured: { context: 'access-requests' } },
+                  { type: 'model', captured: { context: 'upload' } },
+                ],
+              },
+            },
+            {
+              from: { type: 'model', captured: { context: 'access-requests' } },
+              allow: { to: { type: 'model', captured: { context: 'ivas' } } },
```
maryam-sheikhi: It is not duplicate as it seems we have a  similar thing in line 156?

src/app/upload/features/user-upload-grants-list/user-upload-grants-list.ts
Cito: Change again in 313aad5 to delay loading even further to token creation time. Before that, we don't check/show the IVA to keep the page uncluttered.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
Cito: They were accessed in testing. But that was bad anyway, so I also changed the tests. Solved via 82efd86 and 62fb166.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
Cito: It's a simple way to debug in production and we already do it like that in several places.

In the long term I agree, we should have a more consistent and cleaner strategy for frontend side logging, maybe a logger service that could suppress logs in production, or maybe even open telemetry. I have put a note into the refactoring document.

For now, I would keep it here like that, we need a general solution.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
Cito: This is a post request, so it emits only once and then completes or errors out. Therefore we don't need an unsubscribe here.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.ts
Cito: Done for consistency in ad79165.

eslint.config.js
Cito: Yes, I accidentally duplicated the rule. Removed in 6a8923f.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.html
```diff
@@ -0,0 +1,145 @@
...
+      </mat-card-content>
+    </mat-card>
+
+    <p class="w-full">
+      You have been granted upload access to this box from
+      {{ grant.valid_from | date: friendlyDateFormat }} to
+      {{ grant.valid_until | date: friendlyDateFormat }}.
+    </p>
+
+    @if (iva() === undefined) {
```
maryam-sheikhi: What if iva() is null?

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.html
```diff
@@ -0,0 +1,145 @@
...
+  <button
+    mat-button
+    (click)="onClose()"
+    data-umami-event="Upload Work Package Dialog Close Clicked"
+  >
+    Close
+  </button>
+  @if (
+    iva() &&
+    iva()!.state === 'Verified' &&
```
maryam-sheikhi: Why we are calling iva() twice? I mean instead of   iva() && iva()!.state === 'Verified'     we can write  iva()?.state === 'Verified'. Is there any specific reason behind it?

maryam-sheikhi (dismissed):

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.html
Cito: That can normally not happen with the current code. But I changed it against a falsyness check now to cover both.

src/app/work-packages/features/upload-work-package-dialog/upload-work-package-dialog.html
Cito: No, it was copied like that from download. Simplified in both places now.

maryam-sheikhi (approved):

PR #300, opened 2026-04-21
Add space to mat-card title.

maryam-sheikhi:


Cito (changes requested):
This looks nice on the account page.

However, since it's a global style, now some other cards have too mush vertical space (for instance, have a look at the detail page in the access request manager).

I now see that the main problem is that some cards wrap the title text inside a h2 and others do not. To support semantic HTML, I think we should use h2 (or maybe h3 for more deeply nested cards) everywhere, consistently. Then we will also get nicer spacing automatically.

However, we still may need some global style to get the right font size. Something like this:

```css
mat-card-title h2 {
  font: inherit;
  font-size: 1.5rem;
}
```

So, can you check that every mat-card-title in the app uses the proper h2/h3 and when viewed in the browser looks well with the above style? And if needed, adapt so that it looks well.

Cito (changes requested):
Much better now. Still two small things:

In the two dialog boxes, there is still tailwind top-margin of the first content element after the title in 3 occurrences that needs to be removed.

Also, I think the title now is too heavy for the dialog pages (too large and bold). It's ok on the pages, but it should be lighter and smaller in the dialogs. I suggest this global style:

```css
mat-dialog-container mat-card-header mat-card-title h2 {
  font: inherit;
  font-size: 1.25rem;
}
```

Cito (approved):
Pushed remaining changes.

PR #301, opened 2026-04-22
Add method to fetch all studies

Cito:
In order to let the data steward select the study for an upload box, we need to fetch the list of all studies.

This PR adds a method `loadStudiesMap` that does this. Unfortunately, the backend does not provide an efficient way to do that, so this is a bit slow. However, since we will switch to a study based backend soon anyway, it does not make sense to add such functionality on the backend side. And since we cache HTTP GET requests, this will be only slow on the first use. Since this is only temporary, we also don't bother to add a resource API facade but use rxjs.

Note: We could fetch the study description lazily only when the study is selected, but we do it eagerly because it contains the list of its datasets and we can use that to optimize the loading, since some studies have many datasets. Alternatively, instead of getting the dataset summary and the study, we could fetch the dataset details (embedded datasets). That might result in less requests, but more transferred data.

This PR also adds a method `filesOfStudy` to load all file names belonging to a study, with the same caveat of inefficiency.

The PR also fixes a small mistake in the type of the dataset summary.

### DIFF
--- a/src/app/metadata/features/dataset-summary/dataset-summary.html
+++ b/src/app/metadata/features/dataset-summary/dataset-summary.html
@@ -49,24 +49,26 @@
             <mat-icon inline fontIcon="book_2"></mat-icon>
           </div>
           <div class="*:mb-0.5">
-            @if(studiesSummary().count > 0) {
+            @if (studiesSummary().count > 0) {
               <p>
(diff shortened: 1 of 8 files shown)


Cito:
Merged this already since I need it urgently for the mapping tool.

PR #311, opened 2026-05-05
Add size limit to upload box creation dialog

Cito:
The backend now has a mandatory upload box size limit field, so we need to support this in the frontend as well.

As discussed with Paul, we specify the size in TiB. In the context of box sizes, SI TB could also make sense, but I think it's better to consistently use the binary scale throughout the application for individual file sizes, accumulated box sizes and quotas. To make this more explicit, and in the spirit of scientific accuracy, I changed the byte size rendering pipe to output the official GiB unit instead of GB etc. (we showed the commonly used but wrong units before).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "data-portal",
-  "version": "2.5.2",
+  "version": "2.5.3",
   "scripts": {
     "ng": "ng",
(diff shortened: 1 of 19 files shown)


SilverLinings89 (approved):
LGTM

PR #327, opened 2026-06-09
Revert to old way of upload token creation (GSI-2403)

Cito:
Revert to the way we created upload tokens before, which has the advantage of showing missing/invalid IVAs.

This is possible since the registry service now allows users to fetch their own upload grants.


SilverLinings89 (approved):
LGTM and good inclusion of the info box.

PR #331, opened 2026-06-10
Fix handling of incomplete files on upload

Cito:
The backend blocks upload box state changes from open to locked when there are incomplete file uploads.

So this needs to be handled on the user-facing path where the user submits an upload box, not in the upload box manager when the state is set to archived, as it was implemented before.

### DIFF
--- a/src/app/upload/features/upload-box-mapping/upload-box-mapping.ts
+++ b/src/app/upload/features/upload-box-mapping/upload-box-mapping.ts
@@ -4,7 +4,6 @@
  * @license Apache-2.0
  */
 
-import { HttpErrorResponse } from '@angular/common/http';
 import {
   ChangeDetectionStrategy,
   Component,
@@ -33,7 +32,7 @@ import { MatSelectModule } from '@angular/material/select';
 import { MatSort, MatSortModule } from '@angular/material/sort';
 import { MatTableDataSource, MatTableModule } from '@angular/material/table';
 import { RouterLink } from '@angular/router';
-import { of, startWith, switchMap } from 'rxjs';
+import { catchError, concatMap, of, startWith, switchMap, throwError } from 'rxjs';
 
 import { EmFile } from '@app/metadata/models/dataset-information';
 import { Study } from '@app/metadata/models/study';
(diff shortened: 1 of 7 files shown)


src/app/upload/features/upload-box-mapping/upload-box-mapping.ts
```diff
@@ -699,91 +698,32 @@ export class UploadBoxMappingComponent implements OnInit {
         study_id: studyId,
         mapping,
       })
+      .pipe(
+        catchError(() => throwError(() => 'mapping' as const)),
+        concatMap(() =>
+          this.#uploadBoxService
+            .archiveUploadBox(box.id, box.version + 1)
```
SilverLinings89, line 705: does the client actually control this value? It feels like this should fully be managed by the backend... not important for now but usually IDs and versions aren't handled by the client to ensure these cannot be edited maliciously...

src/app/upload/features/upload-box-mapping/upload-box-mapping.ts
```diff
@@ -699,91 +698,32 @@ export class UploadBoxMappingComponent implements OnInit {
...
-          this.#confirmForceArchival(box, incompleteUploads.length);
-        } else {
+        error: (phase: 'mapping' | 'archival') => {
           this.isSubmitting.set(false);
           this.#notificationService.showError(
-            'Mapping was submitted but archival failed.',
+            phase === 'archival'
+              ? 'Mapping was submitted but archival failed.'
+              : 'Failed to submit the file mapping.',
           );
```
SilverLinings89, line 724: I see the issue that people won't know what to do. Also not something we can fix easily but there is no call to action or explanation what to do in either case here.

src/app/upload/features/upload-box-mapping/upload-box-mapping.ts
Cito: If we send the wrong version, the update simply fails. This is to avoid a round-trip fetching the new version because we do two steps at once - kind of optimistic concurrency. The only assumption we make is that new versions are created by incrementing the old one.

src/app/upload/features/upload-box-mapping/upload-box-mapping.ts
Cito: The main error cause are the incomplete uploads in the user-facing path. Most other "anticipated" errors should be excluded by the frontend (unauthorized, wrong state change etc.).

But yes, there is room for improvement in the next release.

PR #332, opened 2026-06-12
Locking and unlocking of boxes by data stewards

Cito:
This PR allows data stewards to lock open boxes and reopen locked ones directly from the "Upload Box Info" card.

### DIFF
--- a/src/app/upload/features/upload-box-manager-detail/upload-box-manager-detail.html
+++ b/src/app/upload/features/upload-box-manager-detail/upload-box-manager-detail.html
@@ -72,6 +72,31 @@
                 {{ box.max_size | parseBytes }}</span
               >
             </p>
+            @if (box.state === 'open') {
+              <button
+                mat-raised-button
+                color="primary"
+                class="mt-4"
+                (click)="lockBox()"
+                [disabled]="isChangingState()"
+                data-umami-event="Upload Box Manager Detail Lock Box Clicked"
+              >
+                <mat-icon fontIcon="lock"></mat-icon>
+                Lock the box
+              </button>
+            } @else if (box.state === 'locked') {
+              <button
+                mat-raised-button
+                color="primary"
+                class="mt-4"
+                (click)="openBox()"
+                [disabled]="isChangingState()"
+                data-umami-event="Upload Box Manager Detail Open Box Clicked"
+              >
+                <mat-icon fontIcon="lock_open"></mat-icon>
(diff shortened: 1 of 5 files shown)


src/app/upload/features/upload-box-manager-detail/upload-box-manager-detail.html
```diff
@@ -72,6 +72,31 @@ <h1>Upload Box Details</h1>
...
+              <button
+                mat-raised-button
+                color="primary"
+                class="mt-4"
+                (click)="openBox()"
+                [disabled]="isChangingState()"
+                data-umami-event="Upload Box Manager Detail Open Box Clicked"
+              >
+                <mat-icon fontIcon="lock_open"></mat-icon>
+                Open the box again
```
SilverLinings89, line 97: Maybe "Unlock the box" for symmetry?

SilverLinings89: Unlock in general - Open to me sounds like "look into" as in open a file.

src/app/upload/features/upload-box-manager-detail/upload-box-manager-detail.ts
```diff
@@ -216,6 +216,127 @@ export class UploadBoxManagerDetailComponent implements OnInit {
...
+    this.#confirmationService.confirm({
+      title: 'Lock upload box?',
+      message:
+        'Locking the box prevents further file uploads by its users. ' +
+        'Are you sure you want to proceed?',
+      confirmText: 'Lock the box',
+      callback: (confirmed) => {
+        if (confirmed) this.#lockBox(box, false);
+      },
+    });
```
SilverLinings89, line 237: We can keep this but honestly, since unlocking is a simple click, too, I personally don't think the extra confirmation is needed here. Unlocking is one click (in the case it's needed) - this way is always one click. Confirmation seems more appropriate when the action is destructive ... Just a thought.

SilverLinings89 (commented):
LGTM in general. Just some very minor questions.

src/app/upload/features/upload-box-manager-detail/upload-box-manager-detail.ts
Cito: You may be right. A data steward should understand the operation and its consequences. But it's a button that is usually not needed in the normal flow.

PR #335, opened 2026-06-22
ES-Lint 10 migration (GSI-2244)

Cito:
- Upgrade to ESLint 10 and cleanup flat-config in eslint.config.js.
- Decouple Prettier from ESLint:
  - drop eslint-plugin-prettier
  - formatting now runs standalone via pnpm format / format:check.
  - eslint-config-prettier/flat is added only to switch off ESLint rules that would conflict with Prettier.
- Replace the unmaintained eslint-plugin-header with a small local rule
- Add .prettierrc overrides (incl. *.stencil.html) and a .prettierignore for build output, generated files, and tooling dirs.
- Pre-commit: .husky/pre-commit now runs pnpm format:check alongside ng lint.
- GitHub lint/test/ci workflows updated for the new commands.
- Re-formatted a handful of HTML templates to match the standalone Prettier pass.


SilverLinings89 (approved):
Good separation.

PR #336, opened 2026-06-24
Fix IVA loading mechanism (GSI-2447)

Cito:
- Resolve the grant's IVA reactively in the download/upload dialogs instead of from a stale snapshot, so it no longer shows a false "Missing IVA" warning on first open.
- Add loading and error states across all IVA views (dialogs, access-grant/upload-grant manager details, user manager) so a slow or failed load no longer reads as a missing IVA or an invalid grant.
- Drop the now redundant eager IVA loading from the access-request button and grants list.
- Standardize inconsistent button sizes and rename misleading CSS classes on the access-request button, dataset summary, and revocation dialogs.


SilverLinings89 (commented):
LGTM

PR #340, opened 2026-06-30
Load file-mapping studies and unmapped files from registry service

Cito:
Replace MetadataSearchService.loadStudiesMap with a new StudyService (placed in the upload section for now until we have a study section) that uses rs to list studies with unmapped files and per-study file-ids, showing only still-unmappable files. We still keep fetching filenames from metldata, since they are not available via rs yet.


SilverLinings89 (commented):
LGTM

PR #349, opened 2026-07-14
Align and improve Dockerfiles (GSI-2421)

Cito:
The DHI image crashed on startup under Kubernetes with `EACCES: permission denied, open 'dist/config.js'`. `config.js` is rewritten on every container start, but it was owned `1000:1000` mode `644`, so it was only writable when the process ran as UID 1000. Clusters that assign an arbitrary UID (e.g. OpenShift) run with GID 0, so the write failed.

Changes:

- Bake `config.js` owned `1000:0` and group-writable (`664`) in both builders, so the write succeeds as UID 1000, as an arbitrary UID with GID 0, or as a pinned `runAsUser: 1000`.
- Align both runner stages: numeric `USER 1000` (drop the custom `appuser`), `WORKDIR /app`, `--chown=${UID}:0` on all copied artifacts.
- Fix pre-existing hadolint warnings (pipefail, consolidated RUNs, documented ignores). `hadolint` now passes clean.

Tested by building both images and running under `docker run` (UID 1000), `--user 99999:0`
(arbitrary UID + GID 0, simulating k8s), and `--user 99999:99999` (negative control, correctly still fails). `config.js` served correctly via `curl` in the first two.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,31 +22,30 @@ RUN jq '{name, version, type}' package.json > ./package.json.run
 # extract js-yaml and its dependencies for runtime
 RUN mkdir -p /tmp/runtime-deps && \
     cp -rL node_modules/js-yaml /tmp/runtime-deps/
+# create config.js, group-writable so it can be written at runtime under any UID
+RUN touch ./dist/data-portal/browser/config.js \
+ && chmod g+w ./dist/data-portal/browser ./dist/data-portal/browser/config.js
 
 # RUNNER: a container to run the service
 FROM base AS runner
-RUN adduser -D appuser
-WORKDIR /home/appuser
-USER appuser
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -6,47 +6,48 @@ RUN apk upgrade --no-cache --available
 
 # BUILDER: a container to build the service dist directory
 FROM base AS builder
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
```
SilverLinings89, line 9: hehe I was so sure that that's a typo :D

SilverLinings89 (approved):

Dockerfile
Cito: Almquist shell :)

PR #350, opened 2026-07-28
Fix issue with saving ticket ID in AR manager (GSI-2524)

Cito:
This PR fixes problems around stale form values read in DOM input handlers.

The FormField directive registers its own 'input' listener after the template listeners, so handlers reading the form model saw the value from before the keystroke, and model writes were overwritten again.

- access request field editor: derive isModified and the ticket ID URL stripping from the form value instead of an (input) handler
- access request duration editor: compare dates by time instead of identity, reset the edit state on cancel and save, and restore only the cleared date to its default
- TOTP and IVA code inputs: sanitize, reset the error and auto-submit  from the form value, removing the reliance on listener order


SilverLinings89 (approved):
LGTM

PR #352, opened 2026-07-29
Support backend pagination for uploaded files (GSI-2494)

Cito:
The RS endpoint `GET /upload-boxes/{id}/uploads` is now paginated and sorted. Uploaded files are therefore paginated on the server instead of in the browser.

This PR assumes the backend gains the ability to sort by accession, and I already created a [PR](https://github.com/ghga-de/ghga-registry-service/pull/25) to add this to RS.

- `UploadBoxService`: `boxFileUploads` is keyed on skip/limit/sort signals and refetches on change; the file table emits page/sort events instead of reordering rows itself.
- Mapping and metadata alignment need the whole box at once, so they use a separate `allBoxFileUploads` resource that walks pages until complete (the endpoint caps `limit` at 1000).
- Deleting a file refetches instead of patching the cached page, which would have shifted the following pages and hidden a file.
- Mocks: `createHandlersForResponses` now derives the requested window from fixtures registered as complete collections, so no fixture per page or sort order is needed.

### DIFF
--- a/src/app/upload/features/upload-box-files-table/upload-box-files-table.ts
+++ b/src/app/upload/features/upload-box-files-table/upload-box-files-table.ts
@@ -27,10 +27,10 @@ import { FileUploadStatePipe } from '@app/upload/pipes/file-upload-state-pipe';
 /**
  * Presentational, paginated table of the file uploads in an upload box.
  *
- * Pagination and sorting are driven by the server: `files` holds a single page
- * as delivered by the RS, and the paginator and sort headers only report the
- * requested page and order via the `page` and `sortChange` outputs. The parent
- * is responsible for fetching the matching page.
+ * Pagination and sorting are driven by the server: `pageFiles` holds a single
+ * page as delivered by the RS, and the paginator and sort headers only report
+ * the requested page and order via the `page` and `sortChange` outputs. The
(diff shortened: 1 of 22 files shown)


src/app/upload/features/upload-box-files-table/upload-box-files-table.ts
```diff
@@ -45,12 +53,27 @@ import { FileUploadStatePipe } from '@app/upload/pipes/file-upload-state-pipe';
   templateUrl: './upload-box-files-table.html',
 })
 export class UploadBoxFilesTableComponent {
-  /** The file uploads to display. */
+  /** The file uploads of the currently shown page. */
   files = input.required<FileUploadWithAccession[]>();
```
SilverLinings89: maybe filesPage would now be more appropriate?

SilverLinings89: or pageContent to be in line with the others.

SilverLinings89 (dismissed):

src/app/upload/features/upload-box-files-table/upload-box-files-table.ts
Cito: I'll use pageFiles as a compromise between the two ideas.

PR #355, opened 2026-08-07
Improve refresh logic and add refresh button (GSI-2509)

Cito:
This PR fixes two reported cases where a browser reload was needed to see current data: after granting yourself an upload box, the account page did not show it; and the "View details" file listing never updated once opened.

Cause: Two independent problems, both hitting each journey:

- **Cached responses were never dropped.** `provideHttpCache({ strategy: 'implicit' })` caches every GET, keyed by URL and query parameters. `HttpCacheManager` was injected in exactly one place in the app, so every other `resource.reload()` silently replayed the cached body. Several existing reloads were no-ops: after submitting an access request, after deleting a file, `reloadUsers()`, `reloadAllIvas()`.
- **Load triggers are idempotent signal sets.** Re-entering a view with the same ID leaves the resource request unchanged, so no request is issued at all. `userGrants` has no trigger of its own and was fetched once per session.

Files are uploaded with the GHGA Connector, outside the portal, so file lists and box sizes change with no portal action. Updating local state cannot keep those current, they have to be fetched again.

Changes:

- Each resource collects its cache keys in a cashew `CacheBucket` via the request context. Dropping a bucket invalidates every variant at once, which the paginated and sorted file uploads need since their keys are not known upfront.
- `reload*` methods on `UploadBoxService`, `AccessRequestService` and `UserService` invalidate before refetching; local-state updates drop their buckets too, so a patched list cannot be resurrected from the cache.
- Views showing data that changes out-of-band fetch again on entry: the account page's upload grants, the upload box details dialog, and the box manager list and detail.
- New shared `app-refresh-button` (`src/app/shared/ui/refresh-button/`), added to all four account page sections, the upload box details dialog, and the upload box, access request, access grant and user manager pages. There was no refresh affordance anywhere before.
- Endpoints whose data changes on its own get a one-minute cache lifetime via `volatileCacheContext()`, as a backstop behind the explicit invalidation. Metadata keeps the long global lifetime, since it only changes with an archive release.
- README and AGENTS.md: document how the test levels relate to the static MSW responses, and state that "end-to-end" is a misnomer for the Playwright tests here.
- Playwright stops after 5 failures on CI instead of 1, so one broken test can be reported for each browser rather than aborting the rest of the run.

### DIFF
--- a/src/app/auth/features/account/account.html
+++ b/src/app/auth/features/account/account.html
@@ -40,7 +40,18 @@
 
 <mat-card appearance="outlined" class="text-primary my-4 p-2">
   <mat-card-header>
-    <mat-card-title><h2>Independent Verification Addresses (IVAs)</h2></mat-card-title>
+    <mat-card-title>
+      <div class="flex items-start justify-between gap-2">
+        <h2>Independent Verification Addresses (IVAs)</h2>
+        <app-refresh-button
+          class="-mt-2"
+          what="your IVAs"
+          [loading]="ivaList.ivasAreLoading()"
+          umamiEvent="User IVA List Refresh Clicked"
+          (refresh)="ivaList.reload()"
+        />
+      </div>
+    </mat-card-title>
     <mat-card-subtitle
       >GHGA uses Independent Verification Addresses (IVAs) to verify your account's
       identity. These IVAs are linked to data access requests and also need to be
@@ -53,13 +64,6 @@
         >GHGA User Documentation</a
       >.</mat-card-subtitle
     >
-    <app-refresh-button
-      class="ml-auto self-start"
-      what="your IVAs"
-      [loading]="ivaList.ivasAreLoading()"
-      umamiEvent="User IVA List Refresh Clicked"
-      (refresh)="ivaList.reload()"
-    />
   </mat-card-header>
   <mat-card-content class="my-4 text-base">
     <app-user-iva-list #ivaList></app-user-iva-list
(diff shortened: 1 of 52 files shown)


src/app/auth/features/account/account.html
```diff
@@ -70,13 +92,24 @@ <h2>
...
 </mat-card>
 
 <mat-card appearance="outlined" class="text-primary my-4 p-2">
   <mat-card-header>
-    <mat-card-title><h2>Pending Access Requests</h2></mat-card-title>
+    <mat-card-title>
+      <div class="flex items-start justify-between gap-2">
+        <h2>Pending Access Requests</h2>
+        <app-refresh-button
+          class="-mt-2"
```
SilverLinings89, line 105: This looks odd, which class did you intend

src/app/auth/features/account/account.html
```diff
@@ -85,13 +118,24 @@ <h2>
...
 </mat-card>
 
 <mat-card appearance="outlined" class="text-primary my-4 p-2">
   <mat-card-header>
-    <mat-card-title><h2>Research Data Upload</h2></mat-card-title>
+    <mat-card-title>
+      <div class="flex items-start justify-between gap-2">
+        <h2>Research Data Upload</h2>
+        <app-refresh-button
+          class="-mt-2"
```
SilverLinings89, line 131: Same here

SilverLinings89 (approved):
Allgemein gut, zwei Anmerkungen zu tailwind Klassen die falsch aussehen. Außerdem: wir sollten kleinere PRs machen, das ist zu viel auf einmal zu reviewen ;)

src/app/auth/features/account/account.html
Cito: `-mt-2` is Tailwind's negative margin-top. It cancels the icon button's padding so the refresh icon lines up with the first line of the heading in the items-start row. Looks a bit better that way.

